PR #1, opened 2018-12-21
search prototype

jpslav:
Has an endpoint for indexing (fixed content) and one for simple searching


jpslav (commented):
This search prototype code will be merged into master prior to productionizing the code.  It's ok.  It is.  Oyez oyez oyez.

PR #5, opened 2019-03-14
skip prototype spec that fails only on travis

jpslav:
Our one and only spec currently fails only on Travis with an error about missing AWS credentials.  The spec uses recorded HTTP interactions from VCR so why this is failing is puzzling.  Since this is prototype code that may or not be in the production implementation, we'll skip it for now so that our Travis runs can complete successfully.


steveburkett:
👍

PR #6, opened 2019-03-14
First Indexing strategy

steveburkett:
First indexing strategy for Openstax books.   (This is a WIP WIP)

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/196


.gitignore
```diff
@@ -11,6 +11,9 @@
 /db/*.sqlite3
 /db/*.sqlite3-journal
 
+# Ignore Rubymine files
+/.idea
```
TomWoodward: editors are a personal choice so i encourage people to add these to their [personal gitignores](https://stackoverflow.com/questions/7335420/global-git-ignore) instead of adding to every project

lib/openstax/page.rb
```diff
@@ -0,0 +1,66 @@
+module Openstax
+  class Page
+    MATCH_PARAGRAPH = '//p[@id]'
+    MATCH_FIGURE = '//figure'
+    MATCH_SELECTOR_PARAGRAPH = 'p'
+    MATCH_SELECTOR_FIGURE = 'figure'
+    MATCH_FIGURE_VISIBLE_TEXT = './/figcaption'
```
TomWoodward: if a figure has text content inside of it, is it always inside a paragraph that would be indexed separately? wondering if there is any content we need to index for figures other than the caption

lib/openstax/page.rb
```diff
@@ -0,0 +1,66 @@
...
+    def process_for_indexable_objects
+      content_dom.xpath(MATCH_ALL_ELEMENTS).each_with_index do | xpath_element, index |
+        indexable_element = create_element(xpath_element, index+1)
+        @indexable_elements << indexable_element if indexable_element
+      end
+    end
+
+    private
+
+    def create_element(ordered_element, page_position)
```
TomWoodward: jp wanted to have all this in one file, but i would prefer put the `create_[element_type]` and `matches?` 
 in separate type objects in a list list so that you can iterate over that and do `matches?` rather than having a conditional in this function for every type

TomWoodward (dismissed):

lib/openstax/page.rb
```diff
@@ -0,0 +1,66 @@
+module Openstax
+  class Page
+    MATCH_PARAGRAPH = '//p[@id]'
+    MATCH_FIGURE = '//figure'
+    MATCH_SELECTOR_PARAGRAPH = 'p'
+    MATCH_SELECTOR_FIGURE = 'figure'
+    MATCH_FIGURE_VISIBLE_TEXT = './/figcaption'
+    MATCH_FIGURE_HIDDEN_TEXT = './/*[@data-alt]'
```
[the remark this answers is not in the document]
steveburkett: yeah..i had these as 2 objects when we were working w/ different types.  can go back to that idea. thanks .

app/controllers/application_controller.rb
```diff
@@ -3,13 +3,7 @@
 class ApplicationController < ActionController::API
 
   def es_client
```
jpslav: I think this method is now unused

app/services/search/book_versions/indexing_service.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def add_book_pages_to_search_index
+      starting = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
+      book = Openstax::Book.new(uuid: @book_guid, version: @book_version)
+      book.pages.each do |page|
+        page.indexable_elements.each do |element|
+          ElasticsearchClient.instance.index(index: index_name,
+                                             type: Search::BookVersions::I1::PageElementType.index_type,
+                                             body: element.to_h)
```
jpslav: I think there's maybe a chance for better encapsulation here -- on line 24 we are hardcoding `I1` which should be part of the strategy injected into this generic class.

```ruby
book.pages.each do |page|
  page.elements.each do |element|
    # The indexing strategy (inside this call) would know which elements it cares about and 
    # can ignore the rest e.g. this one could choose to index paragraphs but skip figures.  It
    # would also be able to supply the correct strategy-specific type information, body, etc.
    # Inside this call is even a place where the page order is recorded (so that if the strategy
    # is handed a paragraph, a figure, and then another paragraph, and it skips the figure,
    # it could give a position of `1` and then `2` for the paragraph index documents.
    indexing_strategy.index(client: Elasticsearch.instance, element: element)
  end
end
```

Having the method on `Page` be `elements` instead of `indexable_elements` would also get the "indexing" language out of `Page` and make it just more purely about the page contents.

jpslav: Moving the specific indexing responsibility into a strategy-specific class also makes the `IndexingService`'s methods operate at more of the same level, which is nice.

lib/openstax/book.rb
```diff
@@ -0,0 +1,44 @@
...
+      @data ||= begin
+        book_json_url = "https://archive.cnx.org/contents/#{uuid}@#{version}.json"
+
+        uri = URI(book_json_url)
+        response = Net::HTTP.get(uri)
+        JSON.parse(response)
+      end
+    end
+
+    def pages
```
jpslav: We will need to handle units in addition to chapters.  Like what goes on here https://github.com/openstax/tutor-server/blob/master/lib/openstax/cnx/v1/book_part.rb#L32-L40 -- book parts can contain other book parts or just pages.

lib/openstax/book.rb
```diff
@@ -0,0 +1,44 @@
...
+          (chapter['contents'] || []).map{|page| page['id']}.compact
+        end
+
+        page_ids.map do |page_id|
+          page_uri = URI("https://archive.cnx.org/contents/#{page_id}.json")
+          page_response = Net::HTTP.get(page_uri)
+
+          begin
+            page = Page.new(id: page_id, data: JSON.parse(page_response))
+            page.process_for_indexable_objects
```
jpslav: see other comment -- would be nice to lazily have the page parse its indexable objects on demand instead of making the call here.

lib/openstax/page.rb
jpslav: I agree with this -- the one file was only nice if it worked out this way, just because it may have been easier when we got to the 2nd and 3rd indexing strategies.

.gitignore
steveburkett: @TomWoodward yup. thanks.  good tip.  

i added this to my ~/.gitconfig, and then inside this file, added my editor's files to exclude.  now i never have to add rubymine to gitignore. i'm stoked! 

[core]
	excludesfile = ~/.local_gitignore

app/controllers/application_controller.rb
steveburkett: @jpslav yeah..there's lots of cleanup we'll be doing as this code evolves.

app/services/search/book_versions/indexing_service.rb
steveburkett: @jpslav i spent yesterday refactoring a lot of this to match up with your and Tom's suggestions.  i'm getting this to work right now....and will then wrap tests around it.  after that we can hook up to the controller plumbing (this i want to chat about) and any other infrastructure logistics.

lib/openstax/book.rb
steveburkett: @jpslav i added a recursive fetch of the page_ids from the TOC.  please check it out to see if it looks correct.  i tried it on the anatomy book....

app/services/search/book_versions/i1/indexing_strategy.rb
```diff
@@ -0,0 +1,57 @@
+module Search::BookVersions::I1
+  # The indexing strategy is the encapsulation for the index's structure and
+  # metdata (including index settings & mappings).
```
jpslav: I see the indexing strategy as mainly the strategy for _how_ to index.  The how is intertwined with the structure since Elasticsearch wants to know what the documents look like.

lib/openstax/book.rb
```diff
@@ -0,0 +1,83 @@
...
+
+          Page.new(id: page_id,
+                   data: JSON.parse(response.body),
+                   desired_page_elements: @indexing_strategy.desired_page_elements)
+        end.compact
+
+        Rails.logger.info "OpenSearch: Loaded book #{@uuid}@#{@version} of #{pages.count} pages"
+        pages
+      end
+    end
```
jpslav: If some of the other code reorg happens and the traversal of books happens more in the indexing strategy, it may make sense to add methods to these book/chapter/page/etc classes that use a visitor pattern, accepting the index strategy to use or ignore parts of the book -- that's a little what's starting to happen already except the strategy is passed during initialization instead of during a `visit` method that crawls an existing tree.

jpslav (changes requested):
Nice progress!

lib/openstax/book.rb
```diff
@@ -0,0 +1,83 @@
...
+      @indexing_strategy = indexing_strategy
+    end
+
+    def pages
+      @pages ||= begin
+        page_ids = recursive_flatten_page_ids(data['tree']['contents'])
+        Rails.logger.info "OpenSearch: found #{page_ids.count} possible page_ids in book #{@uuid}@#{@version}"
+
+        pages = page_ids.map do |page_id|
+          response = fetch_cnx_url(page_uri(page_id)) rescue next
```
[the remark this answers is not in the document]
steveburkett: @jpslav we went thru a lot of this at my last job.  best ways to handle errors like this.  imo, retries usually is a waste of time and effort and messing the code up (if something fails, it's likely it'll fail again).   retries have a place (e.g., in the financial world of transactions) but here...my 2 cent is that we should record the exact information we need, report it to honeybadger (or whatever we use), and move on.  missing out on a page is not the end of the world, better to be sure to get the rest.  And have that information handy for us in excception reporting so that we can see number of times it's happened and relevant page information so we can debug/fix.

lib/openstax/figure_element.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def create_figure(figure_match, page_position)
+      visible_content = figure_match.xpath(MATCH_FIGURE_VISIBLE_TEXT).first.try(:text)
+      hidden_content = figure_match.xpath(MATCH_FIGURE_HIDDEN_TEXT).first.try(:text)
+
+      IndexableElement.new(type: IndexableElement::ELEMENT_TYPE_FIGURE,
+                           page_id: @id,
+                           page_position: page_position,
+                           visible_content: visible_content,
+                           hidden_content: hidden_content)
```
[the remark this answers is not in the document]
steveburkett: @jpslav my idea on this was that this object encapsulates the figure element content.  If there's a future one, we'd create another object and assign that object to the new indexing strategy.  

i thought it made sense to SRP these objects...parsing these object is their responsibility.  We can add new ones as a point of extensibility.  And then assign the new ones, or different ones, to the various strategies.

app/services/search/book_versions/indexing_service.rb
```diff
@@ -0,0 +1,67 @@
...
+    def create_index
+      ElasticsearchClient.instance.indices.create(index: index_name, body: @indexing_strategy.index_metadata)
+    rescue => ex
+      Rails.logger.error "OpenSearch (error): unable to create index for book #{index_name}: #{ex.message}"
+    end
+
+    # A book's page data is loaded when book.pages is called
+    # A page is parsed into indexable elements when page.indexable_elements is called
+    # TODO: sending exceptions to exception service, where to continue where to fail
+    def add_book_pages_to_search_index
```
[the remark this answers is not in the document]
steveburkett: @jpslav i renamed it to 'populate'.  this sounds more exact and simple.

app/services/search/book_versions/indexing_service.rb
```diff
@@ -0,0 +1,67 @@
...
+      starting = Time.now
+
+      book = Openstax::Book.new(uuid: @book_guid,
+                                version: @book_version,
+                                indexing_strategy: @indexing_strategy)
+      book.pages.each do |page|
+        page.indexable_elements.each do |element|
+          indexing_strategy.index_row(element: element, index_name: index_name)
+        end
+      end
```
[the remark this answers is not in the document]
steveburkett: @jpslav if a new requirement came in to add more elements (like a unit and chapter name), then my thought there was we'd create these element objects (UnitElement, ChapterElement) & add these to the new strategy.   Then, when Search::BookVersions::Index#populate is called, the Book's parsing would use this new strategy which declares that it's interested in UnitElement, ChapterElement, and FigureElement)....  these would be the ones parsed out of the book in I2 (as IndexableElements).   The 'populate' code ^ doesnt care much about the indexing strategy..it just blissfully finds all the "IndexableElements" and sends them to ElasticSearch.

spec/services/search/book_versions/index_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    before do
+      allow(OpenStax::Cnx::V1).to receive(:fetch).with(test_book_url).and_return(test_book_json)
+      allow(OpenStax::Cnx::V1).to receive(:fetch).with(test_page_url).and_call_original
+    end
+
+    it 'populates the index' do
+      index.create
+      index.populate
+      sleep 1 # a necessary evil to get the real elastic search document count
```
jpslav: Let's only sleep if we are recording the cassette:

```ruby
sleep 1 if VCR.current_cassette.try(:recording?) # to get the real elasticsearch document count
```

jpslav (changes requested):
As discussed please delete the blank specs that are now the job of cnx-ruby

spec/services/search/book_versions/index_spec.rb
steveburkett: @jpslav  it only works w/ the sleep 1 in there either recording or not.

jpslav (approved):

lib/openstax/page.rb
jpslav: @philschatz do you know the answer to this? ^^

PR #10, opened 2019-03-29
add signed requests to Elasticsearch

jpslav:
* Moves Elasticsearch client wrapper to `/lib/open_search`
* Adds code to sign requests with AWS credentials for communicating with the ES cluster securely.  
* Adds a recoded spec showing these AWS requests succeeding with signatures and failing without.


lib/elasticsearch_client.rb
```diff
@@ -0,0 +1,55 @@
+require 'elasticsearch'
+
+class ElasticsearchClient
```
steveburkett: my preference would be that we name this to be more specific to us.  it's not really a ES client (that's in the Elasticsearch ruby integration gem).  it's OSElasticSearchClient (this helps to state that it's our internal one and that we use the Elasticsearch::Client (it's clear that this is the gem's version).

steveburkett: Also, all this could be a bit simpler if you want.... e.g., this is what i did in my other project:

`initializers/elasticsearch.rb`
```
require 'faraday_middleware/aws_signers_v4'

elasticsearch_params = {
  log: true,
  url: ENV.fetch('AWS_ELASTICSEARCH_URL')
}

OpenSearch::ElasticSearchClient = Elasticsearch::Client.new(elasticsearch_params) do |f|
  f.request(
    :aws_signers_v4,
    credentials: Aws::Credentials.new(ENV.fetch('AWS_ACCESS_KEY'), ENV.fetch('AWS_SECRET_ACCESS_KEY')),
    service_name: 'es',
    region: ENV.fetch('AWS_REGION')
  )
end
```

spec/lib/elasticsearch_client_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      signing_client = ElasticsearchClient.new(
+        url: "https://#{domain_status.endpoint}",
+        sign_aws_requests: true
+      )
+
+      resp = signing_client.search q: "test"
+      expect(resp).to have_key("took")
+
+      non_signing_client = ElasticsearchClient.new(
```
steveburkett: this might be clearer specified as a named subject.

spec/lib/elasticsearch_client_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        url: "https://#{domain_status.endpoint}",
+        sign_aws_requests: false
+      )
+
+      expect {
+        non_signing_client.search q: "test"
+      }.to raise_error(Elasticsearch::Transport::Transport::Errors::Forbidden)
+    end
+  end
+
```
steveburkett: @jpslav  do you care one way or the other about blank lines before and after class declarations/ends?  i see this pattern in your code.  however, i dont see this pattern in the ruby coding standards https://github.com/rubocop-hq/ruby-style-guide#indent-public-private-protected (it has suggestions on blank lines, but not just after class declarations/ends.  Asking if you care and if there's a pattern you want us to follow.  My 2 cents is to follow the ruby rubocop standards on this (which is no blank lines in these specific spots).

spec/vcr_helper.rb
```diff
@@ -46,3 +47,23 @@
...
+# in before hooks.  Since there isn't a way to make sure that the content
+# length update happens after filtering senstive data (and since filtering
+# data can mess with the content length), we end up not getting the benefit
+# out of the update_content_length_header hook that we are shooting for.
+# This monkey patch fixes this by forcing a content length update after
+# every filtering of sensitive data.  This does more work than we need
+# but it is also fast so it isn't that big of a deal.
+
+class VCR::HTTPInteraction::HookAware
+  alias_method :orig_filter!, :filter!
```
steveburkett: here..you're not using blank line after class and before class end.  so this seems a bit more normal to me.  but just let me know ;)

lib/elasticsearch_client.rb
jpslav: Making it `OpenSearch::ElasticsearchClient` works for me.

I don’t know if we can simplify the code quite this much because we have three kinds of configurations to handle: 1) local development where ES is in a docker container and requests aren’t signed, 2) test development where we sign requests using key and secret like you have above, and 3) production use where we use the built-in credentials on EC2 instances.  The wrapper class in this PR lets us do all three of these and lets us control the endpoint via a secret.  Does you agree?

Incidentally, I saw the client definition in your PR was in `config/initializers` but I have not typically put class definitions in that folder (instead used it for configuration calls), which is why I put this one in `lib`.  

(separately I thought I had to use `patron` is that not your experience?)

spec/lib/elasticsearch_client_spec.rb
jpslav: My preference is to not enforce a rule here.  I probably do it both ways and I think I choose in the moment based on what looks clearer to me.  Here I wanted the before block to have a little more visual separation.  In the later example with the monkey patch, it was very short and I went with a compact look.

spec/lib/elasticsearch_client_spec.rb
jpslav: I tend to find uses of `subject` (the implicit and unnamed kinds, which I realize is not what you’re talking about) confusing, and as such normally avoid it for all of its variants.  I like this [SO answer](https://stackoverflow.com/a/38459039) I just came across, as your comment made me think about this! I think instead of a named subject I’d just use a `let`, and then we wouldn’t have uses of the anonymous subject which could be awkward when there are potentially multiple subjects as there would be here. 

There are some times I have used an anonymous subject, like when I have many little tests to write about something and it is clear in the spec what the subject is.

steveburkett: agree on the use of an implicit subject.  I personally like named subjects (they're pretty much the same thing as a let, but with a tiny bit of extra meaning in that it makes it obvious what your subject is for the test).  this seems like a minor point tho.  Not a big deal imo.  I just tend to like the code to be as obvious as possible...and a bit less rails/ruby magic as possible.  

Going off on a tangent here....Sandi Metz gave us a good point when we met her:  dont be afraid of un-DRYness in your tests, and to lean more toward being descriptive and explicit (over clear and DRY).  DRYness for concepts in the code, but not DRYness just for the sake of dryness. (This was a great meeting with her.  Would be cool of OpenStax can get her in for a week somewhere for all devs!)

lib/elasticsearch_client.rb
steveburkett: class definitions in initializers are fine, or the global i used ^.  Not sure it matters that much. I just like that stuff that is "initialized" is there (perhaps it's using a class defined in lib).   I tend to think of things in lib as stuff that you would eventually convert to a gem (possibly).   

the wrapper class you wrote is good.  i just tossed this out there in case to consider a simpler idea, if possible...but it sounds like not the case. 

I've used the above ^ code in my other project working with ES both locally (unsigned, locally installed) and in production (signed, AWS ES).  it works for both (the local version just ignores the credentials, and uses a localhost:9200 url)

PR #12, opened 2019-03-30
enable optional 'open-search' URL path prefix

jpslav:
This PR builds off of @KarinaMendez2's work on Accounts to add an `/accounts` path prefix.  I put her code into a new middleware gem (https://github.com/openstax/path_prefixer) and used it here to give an `/open-search` optional path prefix.  This will let this search code be used under a CloudFront distribution that shares a top-level domain across multiple properties.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,8 @@ gem "whenever"
 # Unlike in most of our projects, this is used in production (to set the node type)
 gem 'dotenv-rails'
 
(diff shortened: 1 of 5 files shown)


KarinaMendez2 (approved):
It works great! In my code I had "if 'GET' == env['REQUEST_METHOD']" because at some point based on the examples we used for the middleware, I thought it would only be needed for `GET` requests but will switch to using your gem now in accounts as soon as we can.

PR #13, opened 2019-04-05
Separating CNX knowledge from indexing knowledge

jpslav:
@steveburkett here's the idea for modifying the indexing PR.  Let's chat next week!

### Main changes

* Changes `lib/openstax` code to contain only knowledge of CNX data structures (nothing about indexing).  CNX data structures are pretty stable, so by limiting this area of the code only to those structures, we can expect not to see many changes here.
* Changes `Search::BookVersions::I1::IndexingStrategy` to know _how_ to index a book in addition to the structure of a book index.  This code and its peers become responsible for traversing a book's structure and indexing whichever components are interesting to the strategy at hand.
* Adds `Document` classes (in the index strategy folder) that wrap `Element` classes (from the CNX book lib folder).  `Document` classes know about "visible" and "hidden" terms.  In our first indexing strategy there is one document type known to ES.  That is defined by the base class `PageElementDocument` (which used to be `PageElementType`) and specific subclasses (`ParagraphDocument`, `FigureDocument`) that override methods in the base class to give element-specific values for hidden and visible fields (e.g. the `FigureDocument` implements its `visible_content` method by calling the `caption` method on its wrapped element (which in turn does has the xpath code to find the text in the caption XML).
* When we add the second and third indexing strategies, changes should be confined to the `Search::BookVersions::I2` and `I3`.
* This PR also suggests that we pull in the existing book parsing code from tutor-server (we should probably pull it into a gem and toss any tutor-specific code).

### Minor TODO

We need to change all "Openstax" uses to "OpenStax" -- we can add an inflection if Rails gets confused looking for "open_stax" folders.

```ruby
ActiveSupport::Inflector.inflections do |inflect|
  inflect.acronym 'OpenStax'
end
```


steveburkett:
@jpslav Was the book page_ids recursive code wrong?  I formulated this after chatting w/ @nathanstitt .  It seemed like a significant simplification over the tutor code (which i didnt really follow all that well, since it seemed like it was scattered around, some of which depended on css data that lived in the database.

lib/openstax/book.rb
```diff
@@ -4,88 +4,101 @@ module Openstax
...
-      if response.code.to_i >= 400
-        Rails.logger.debug "OpenSearch (error): book (#{@uuid}@#{@version}), uri #{uri}, error code: #{response.code}"
-        raise "OpenSearch Unable to find Uri #{uri} #{response.code}"
-      end
-
-      response
-    end
+    # ******************************************************************************************
+    #
+    # Let's consider using the existing book parsing logic in tutor-server with the tutor-
```
steveburkett: @jpslav i totally support reusing code for sure.  but in this case, after looking at the tutor code (and discussing w/ @nathanstitt) i thought that the recursive code I wrote here was a more direct path to the desired target of pages (which is what we're interested in for indexing, at least for I1).   The tutor code, quite honestly, i couldnt follow (easily)... It's pretty complicated and relies on data structures (css selectors) stored separately in the database to navigate page structure (I think).  If we need to do this, which it sounds like we do need to do, it would be helpful to have some pairing help to extract a clean, self reliant version of the tutor code.

jpslav: I think the part of the Tutor code you and Nathan are referring to has to do with how Tutor splits pages into "fragments" (e.g. it may chop a single page into 3 fragments).  That code does use "processing instructions" that live in Tutor and tell the CNX code how to split the pages based on CSS it finds within the page.  **That code is Tutor-specific and _not_ something we or a general client would need.**  I do totally understand your desire to make something simpler after looking at that code!

There was nothing wrong with your recursive code to get pages.  But when we move indexing logic out of that code, what is left becomes similar to a big chunk of the existing (spec- and time-tested) CNX code in Tutor.  So I'd rather build on that chunk if possible.

I just took the Tutor CNX code, removed the fragment code, and put the remainder into a new `openstax_cnx` gem: https://github.com/openstax/cnx-ruby.  I filled out the README with examples of how to traverse book contents, so take a gander and see if it looks promising.

If the new gem looks promising, we can move the `FigureElement`, `ParagraphElement`, and `Page.elements` into it.  This would put all of the CNX knowledge (and CNX processing specs) in one place.  We could also then consider extending the gem to provide factories for building test books and pages for use by open-search.

Let me know what you think.

app/services/search/book_versions/i1/figure_document.rb
```diff
@@ -0,0 +1,20 @@
+module Search::BookVersions::I1
+  class FigureDocument < PageElementDocument
+
+    def initialize(element:, page_position:, page_id:)
```
steveburkett: maybe "element" could be more specifically named?   openstax_element, xpath_element, or something...to make it clearer.

app/services/search/book_versions/i1/indexing_strategy.rb
```diff
@@ -8,6 +8,11 @@ class IndexingStrategy
     NUM_SHARDS = 1
     NUM_REPLICAS = 1
 
+    DESIRED_ELEMENTS_TO_DOCUMENTS = [
```
steveburkett: did you mean a hash here?

PR #14, opened 2019-04-07
Fix RexReleases spec on Travis

jpslav:
The RexReleases spec had been failing only on Travis, and so we disabled it a few weeks ago.  I had a thought about how to fix it -- basically to set dummy AWS secrets in the `ENV` even though they aren't needed since we are using recorded interactions.  I created this PR to test this out, but when I reenabled the spec it passed in Travis with no other modifications... so yay I guess :-)


jpslav:
@steveburkett - merged without review as this PR just reenables one existing spec.

PR #18, opened 2019-04-10
longer persistent connection timeout

jpslav:
CloudFront lets us set a persistent connection timeout up to 60 seconds.  To prevent Puma from closing the connection before then, this PR ups its persistent connection timeout to 62 seconds.


jpslav:
Tiny, merging without review.

PR #20, opened 2019-04-13
script to run ES for specs

jpslav:
A script to run Elasticsearch from Docker for development and spec purposes


jpslav:
fyi @steveburkett

PR #22, opened 2019-04-15
Initial creation of enqueue objects

steveburkett:
Create the rake task that reads rex releases for book versions to be indexed.   For each book in the rex release, create a dynamo entry for it and enqueue a SQS job.   

The DynamoDB entry records the book version metadata to be indexed.  

The SQS job is the job to be executed for indexing by a worker.  

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/197


app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,36 @@
+# Indexes a book version thru the SQS queuing system
+#
+# @params indexing_version e.g., Search::BookVersions::I1::IndexingStrategy::VERSION
+class EnqueueIndexJobs
```
jpslav: My reading of this class was that it contained all the work for the rake script, so the rake script could just call off to this PORO.  But looking at how it is initialized with a single indexing strategy I'm getting the feeling that may not be the case?

If this is intended to be just for one indexing strategy, that means that when running for multiple indexing strategies we'll be repulling and iterating through all the releases multiple times.  I'm not sure if that is the best.  

Above is discussed how the validation approach to checking if a BookIndexing already exists might not be ideal.  Also with our desire to delete no-longer-requested book/indexing strategy combos, an all-at-once approach like the following might be better (the below is very rough):

```ruby
class EnqueueIndexJobs
  def do_it
    rex_releases = OpenStax::RexReleases.new
    book_ids = rex_releases.map{ |release| release.books }.flatten

    book_indexings = BookIndexing.all # guessing that this is how you get all entries from Dynamo

    book_ids.each do |book_id|
      active_indexing_strategies.each do |indexing_strategy|
        # to avoid iterating through the entire book_indexings array over and over, probably good
        # to make a hash of book IDs to hashes of indexing strategies to BookIndexings

        book_indexing = that_crazy_hash[book_id].try(:[], indexing_strategy)

        if book_indexing.nil?
          # hey, we haven't done this one yet! 
          job = IndexingJob.new(book_version_id: ..., indexing_strategy: ...)
          @todo_jobs_queue.write(job) 
          BookIndexing.create!(...)
        else
          # we have done this one already, somehow note that it is still in demand
          # ones we don't mark don't have a book ID / indexing strategy combination
          # indicated in the released content, and should be deleted (below)
        end
    end

    # find all of our local book indexings that we didn't mark in demand and delete 
    # all of their indexes

  end
end
```

app/models/book_indexing.rb
```diff
@@ -0,0 +1,57 @@
...
+  # the completion a "job" in the SQS indexing pipeline
+  def finish(book_version_id:)
+    # TODO
+  end
+
+  private
+
+  def book_indexing_version_is_unique
+    found_books = BookIndexing.where(book_version_id: book_version_id,
+                                     indexing_version: indexing_version)
```
jpslav: I'm pretty sure this is not an efficient way to operate.  DynamoDB is a simple key value store and can't do efficient queries like a RDBMS.  Looking at the original Dynamoid blog post, a `where` query "does not efficiently perform a conditional join, but instead finds all the user's addresses and naively filters them in Ruby".  So combined with the loop below in `EnqueueIndexJobs` this means we are querying all of the book indexing records over and over.  I suspect we'll want to find a different approach.

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,36 @@
...
+                                        indexing_version: @indexing_version)
+
+    TodoJobsQueue.new(book_version_id: book_id,
+                                 indexing_version: @indexing_version).write
+
+  rescue => ex
+    book_indexing.destroy if book_indexing
+
+    if ex.is_a?(Dynamoid::Errors::DocumentNotValid)
+      Rails.logger.info "Book version #{book_id} already being indexed by IndexingJob"
```
jpslav: In fact it may have been indexed weeks ago, so we don't want to log this (would be a lot of duplicated log messages for what is an expected situation)

jpslav (changes requested):
I did not review the specs but I do like that there are a lot of them!

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,70 @@
...
+    rex_releases = OpenStax::RexReleases.new
+    books_ids = rex_releases.map{ |release| release.books }.flatten
+
+    books_ids.each do |book_id|
+      ACTIVE_INDEXING_STRATEGIES.each do |indexing_strategy|
+        process_book(book_id, indexing_strategy.new.version)
+      end
+    end
+
+    check_for_books_to_delete
```
jpslav: I understand the desire to put code into private methods, but here it doesn't help me read this code and understand what's going on.  "process_book" is kind of vague.  Also, `process_book` sets state that is used in `check_for_books_to_delete` so for that I'd rather just have the code more inline here.  I made a PR into this PR to show you what I mean - it ends up as 24 lines of code but keeps logic together and is more readable to me (note that that PR is pseudocode-y so you may not just be able to merge it).

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    AutoScaling.set_desired_capacity(group_name: asg_name, desired_capacity: @todo_jobs_queue.count)
```
jpslav: In that PR I changed this to just an ASG instance (more instance methods, less class methods)

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,70 @@
...
+    books_to_be_deleted.each do ||
+      job = DeletingJob.new(book_version_id: book_id, indexing_version: indexing_version)
+      enqueue_book(job, BookIndexing::DELETE_PENDING)
+    end
+  end
+
+  def find_book_in_indexing(book_id, indexing_version)
+    book_indexings.detect do |book_indexing|
+      book_indexing.book_version_id == book_id &&
+         book_indexing.indexing_version == indexing_version
```
jpslav: This could be pretty inefficient -- iterating through 1500 entries for each of 1500 combinations of inputs; O(N^2).  In my PR I cached a hash of the book indexings (O(N)) to make lookup O(1).

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def enqueue_book(job, state)
+    @todo_jobs_queue.write(job)
+
+    BookIndexing.create(book_version_id: book_id, indexing_version: indexing_version, state: state)
+
+    Rails.logger.info "Open-Search: Book version #{book_id} #{indexing_version} loaded into todo_job_queue"
+  end
```
jpslav: in PR switched these to `enqueue_create_index_job` and `enqueue_delete_index_job`.  Even if don't use these words, prefer "enqueing jobs" rather than "enqueueing books" b/c we aren't really queuing books.

config/initializers/indexing_strategies.rb
```diff
@@ -0,0 +1,7 @@
+ACTIVE_INDEXING_STRATEGIES = [
+  Search::BookVersions::I1::IndexingStrategy
```
jpslav: could put the strings here instead of classes, so don't have to say `.new.version` above which may not be same for all strategies

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,70 @@
+# Indexes a book version thru the SQS queuing system
+#
+# @params indexing_version e.g., Search::BookVersions::I1::IndexingStrategy::VERSION
+class EnqueueIndexJobs
+  def initialize
+    @todo_jobs_queue = TodoJobsQueue.new
+  end
+
+  def process_rex_releases
```
jpslav: changed name to `execute` in the PR.  B/c does an outside caller really care that this starts by pulling rex releases?  I think this Ruby PORO is still just a script but now one that has attributes/state.  So I like just saying `do_it` or `call` or similar.

app/models/book_indexing.rb
```diff
@@ -0,0 +1,59 @@
+class BookIndexing
+  include Dynamoid::Document
+
+  VALID_STATES = [
```
jpslav: These are all the states, so why prefix with "VALID_".  Is there an "invalid" state?  Maybe just call "STATES"

app/models/book_indexing.rb
```diff
@@ -0,0 +1,59 @@
...
+    BookIndexing.new.tap do |job|
+      job.state = state
+      job.book_version_id = book_version_id
+      job.indexing_version = indexing_version
+      job.enqueued_time = DateTime.now
+      job.save!
+    end
+  end
+
+  def self.valid_book_indexings
```
jpslav: This use of "valid" doesn't match "VALID_STATES" above.  Also, I don't think a "deleted" state is invalid per se.

app/models/todo_jobs_queue.rb
```diff
@@ -0,0 +1,14 @@
...
+    super(url)
+  end
+
+  def write(indexing_job)
+    raw_queue.send_message( message_body: indexing_job.to_json )
+  end
+
+  def read
+    message = raw_queue.receive_messages()
+    message.size == 0 ? nil : IndexingJob.new(message)
```
jpslav: You can comment this out until working on worker code if you want

jpslav (changes requested):

app/services/enqueue_index_jobs.rb
steveburkett: this is just my opinion...so would love to discuss more...but i've disliked using "meaningless" method names like execute or do_it or call (in services).  it adds nothing to the code documentation and understanding, when what is it "calling" or "doing' is adding to easier code understanding.  method names are the source for documenting what the code or method is doing, hopefully a single responsibility.

app/services/enqueue_index_jobs.rb
steveburkett: this iterates thru the book indexings just once and then memoizes the results.

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def check_for_books_to_delete
+    books_to_be_deleted = book_indexings.select do |book_indexing|
+      book_indexing.in_demand == false
+    end
+
+    books_to_be_deleted.each do ||
+      job = DeletingJob.new(book_version_id: book_id, indexing_version: indexing_version)
```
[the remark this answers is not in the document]
steveburkett: yeah...agreed.  i thought the naming of this value object could be better too....

app/services/enqueue_index_jobs.rb
jpslav: this method is what could be the inlined body of the rake script, so to me the name of this method is really "enqueue_index_jobs".  of course that is also the class name here, which is why a "do_it" method makes sense to me (because `EnqueueIndexJobs.new.enqueue_index_jobs` is repetitive).  Said another way, `EnqueueIndexJobs` is a glorified function (a function that is an object so that it can have state), and functions are `call`ed.  Does `process_rex_releases` help here with the understanding of what's going on?  If a new user came to this code, would it be obvious that `process_rex_releases` is what should be called to run this "enqueue index jobs" function?  To me, it isn't really obvious other than it being the only public method.

app/models/book_indexing.rb
steveburkett: @jpslav do you like the states prefixed like this? 
```
  STATES = [
    STATE_PENDING = "pending",
    STATE_DELETE_PENDING = "delete pending",
    STATE_STARTED = "started",
    STATE_INDEXED = "indexed",
    STATE_DELETED = "deleted"
  ]
```

app/services/enqueue_index_jobs.rb
steveburkett: thanks @jpslav.  We argued over this for hours at my previous job.... i'm cool with this and agree...that the idea of the service object is that of a glorified function object, to invoke call or execute on. 

i do hope tho that we dont pack our (production) rake scripts with functional code.  same reasoning as packing a controller with functional code (harder to test, less SRP, etc).

jpslav (changes requested):

app/services/enqueue_index_jobs.rb
jpslav: this old review comment appears to be disconnected from its original location?

app/models/book_indexing.rb
```diff
@@ -0,0 +1,72 @@
+class BookIndexing
+  include Dynamoid::Document
+
+  table name: Rails.application.secrets.dynamodb[:index_state_table_name].parameterize.underscore.to_sym
+
+  STATES = [
+    STATE_PENDING = "pending",
+    STATE_DELETE_PENDING = "delete pending",
+    STATE_STARTED = "started",
```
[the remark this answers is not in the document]
steveburkett: @jpslav i just guessed at this...i think it's what we'd set the dynamo object to when this job is started.  does that sound right?   (and then indexed when finished).

app/models/book_indexing.rb
```diff
@@ -0,0 +1,72 @@
+class BookIndexing
```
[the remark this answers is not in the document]
steveburkett: if we rename this, my preference would be something like BookIndexingJob or the like.  because it mostly accurately represents a job (in the domain world)...it contains state and a few other things (like timestamps).  

i do agree we should strive toward the best naming we can muster here...the word indexing is kind overloaded.

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,87 @@
...
+  def initialize
+    @todo_jobs_queue = TodoJobsQueue.new
+    @new_create_index_jobs = 0
+    @new_delete_index_jobs = 0
+    @worker_asg = AutoScalingGroup.new(worker_asg_name)
+  end
+
+  def call
+    released_book_ids.each do |book_id|
+      ACTIVE_INDEXING_STRATEGIES.each do |indexing_version|
```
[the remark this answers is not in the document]
steveburkett: i used indexing_version because that what it was.   a strategy is the indexing object/class.  agree we should be consistent.  are u ok with calling this ACTIVE_INDEXING_VERSIONS?

app/models/book_indexing.rb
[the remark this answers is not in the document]
steveburkett: @jpslav and I chatted more on this.   it was a miscommunication on my part that this class should be a state of indexing, and not a "activity log".  so...will rename to JP's recommendation.

app/models/book_indexing.rb
```diff
@@ -0,0 +1,72 @@
...
+  field :indexing_version
+  field :enqueued_time, :datetime, store_as_string: true
+  field :started_time,  :datetime, store_as_string: true
+  field :finished_time, :datetime, store_as_string: true
+  field :message
+
+  attr_reader :in_demand
+
+  def self.create_new_indexing(book_version_id:, indexing_version:)
+    Rails.logger.info "Creating book version #{@book_version_id} #{@indexing_version} in dynamodb"
```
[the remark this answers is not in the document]
steveburkett: i can use less logging.  I'm used to a shop where we logged everything (because it was hard to figure out what was wrong in the system).

app/models/book_indexing.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def deleting?
+    [STATE_DELETED, STATE_DELETE_PENDING].include?(self.state)
+  end
+
+  def queue_to_delete
+    self.state = STATE_DELETE_PENDING
+    self.enqueued_time = DateTime.now
+    save!
+  end
```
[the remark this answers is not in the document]
steveburkett: absolutely!  good catch.

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    Rails.logger.info "Open-Search: Book version #{book_id} #{indexing_version} enqueued for indexing"
+  end
+
+  def enqueue_delete_index_job(book_indexing)
+    job = DeleteIndexJob.new(book_version_id: book_indexing.book_version_id,
+                          indexing_version: book_indexing.indexing_version)
+    @todo_jobs_queue.write(job)
+
+    book_indexing.queue_to_delete
```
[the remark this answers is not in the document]
steveburkett: mark_queued_for_deletion is a more accurate name

spec/support/sqs.rb
```diff
@@ -0,0 +1,17 @@
+module Sqs
+  def self.create_test_queue
+    random_queue_name = "test-#{(0...8).map { (65 + rand(26)).chr }.join}"
```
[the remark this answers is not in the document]
steveburkett: using a simple constant test queue name...dont see the need for a random name....

spec/models/book_indexing_spec.rb
```diff
@@ -1,9 +1,13 @@
 require 'rails_helper'
 require 'vcr_helper'
 
+amazon_api_header_matcher = lambda do |request_1, request_2|
```
steveburkett: great hackery!

jpslav (approved):

app/models/book_index_state.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def initialize(*args)
+    super
+    self.in_demand = false
+  end
+
+  def mark_queued_for_deletion
+    self.state = STATE_DELETE_PENDING
+    self.enqueued_time = DateTime.now
```
jpslav: What time zone does this get set in?  Brief looking around in Dynamoid suggests server's local time. Should we standardize to UTC?

app/models/book_index_state.rb
```diff
@@ -0,0 +1,74 @@
...
+      job.book_version_id = book_version_id
+      job.indexing_version = indexing_version
+      job.enqueued_time = DateTime.now
+      job.save!
+    end
+  end
+
+  def self.live_book_indexings
+    all.reject{ |doc| doc.deleting? }
+  end
```
jpslav: Change name away from `book_indexings` to match name change of class.  Also here `live` could work

app/services/enqueue_index_jobs.rb
```diff
@@ -0,0 +1,94 @@
...
+  def stats
+    {
+      num_delete_index_jobs: @new_delete_index_jobs,
+      num_create_index_jobs: @new_create_index_jobs
+    }
+  end
+
+  def book_indexings
+    @book_indexings ||= BookIndexState.live_book_indexings
+  end
```
jpslav: Change name of this method/variable and other uses in this class away from "BookIndexing" to match name change to BookIndexState, e.g. `def index_states`

jpslav (changes requested):

app/models/book_index_state.rb
```diff
@@ -0,0 +1,74 @@
...
+        key: :book_version_id
+
+  range :indexing_version
+
+  field :state
+  field :enqueued_time, :datetime, store_as_string: true
+  field :started_time,  :datetime, store_as_string: true
+  field :finished_time, :datetime, store_as_string: true
+  field :updated_at,    :datetime, store_as_string: true
+  field :created_at,    :datetime, store_as_string: true
```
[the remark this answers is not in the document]
steveburkett: @jpslav i put these here so we could get store_as_string = true instead of the unix time, which is easier to grok

app/models/book_index_state.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  table name: Rails.application.secrets.dynamodb[:index_state_table_name].parameterize.underscore.to_sym,
+        key: :book_version_id
+
+  range :indexing_version
+
+  field :state
+  field :enqueued_time, :datetime, store_as_string: true
+  field :started_time,  :datetime, store_as_string: true
+  field :finished_time, :datetime, store_as_string: true
```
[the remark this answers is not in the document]
steveburkett: in #198, we'll move to a status_log hash, replacing the timestamps

app/models/book_index_state.rb
steveburkett: for sure we should use UTC (will test this now to confirm what dynamoid stores)

steveburkett: yeah, dynamoid saves as UTC time

PR #24, opened 2019-05-09
Worker processes the SQS queue

steveburkett:
Add a rake task and functionality to process the enqueued indexing and delete-index jobs. Once the jobs are processed, they are moved to the Done SQS queue where they will be marked as done by the manager instance.  

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/198


app/services/search/book_versions/index.rb
```diff
@@ -39,14 +38,25 @@ def recreate
     end
 
     def delete
-      OpenSearch::ElasticsearchClient.instance.indices.delete(index: name)
+      if OpenSearch::ElasticsearchClient.instance.indices.exists?(index: name)
+        OpenSearch::ElasticsearchClient.instance.indices.delete(index: name)
```
jpslav: Might dry up code a bit to add...

```ruby
def indices
  OpenSearch::ElasticsearchClient.instance.indices
end
```

lib/open_search/elasticsearch_client.rb
```diff
@@ -16,26 +16,44 @@ def self.instance
...
+
+        f.adapter :typhoeus
       end
     end
   end
 
   protected
 
+  # setting the faraday timeout to 15 seconds here.  it times out very often
+  # if using the default of 1.5 seconds.
```
jpslav: Is this still true after the switch to `typhoeus`?

app/services/search/book_versions/index.rb
```diff
@@ -39,14 +38,25 @@ def recreate
     end
 
     def delete
-      OpenSearch::ElasticsearchClient.instance.indices.delete(index: name)
+      if OpenSearch::ElasticsearchClient.instance.indices.exists?(index: name)
```
jpslav: What happens if doesn't exist and try to delete it?  If it doesn't do anything bad, we can take this `exists?` call out.  If it does raise an exception, let's change to making the call and rescuing the exception, because as is there's a very-rare chance that the index could be deleted between the `exists?` and `delete` calls anyway.

lib/tasks/work_index_job.rake
```diff
@@ -5,6 +5,8 @@ task work_index_job: :environment do
 
   Rails.logger.info { "Ran placeholder work_index_job task!" }
 
+  ProcessIndexJobs.new.call
```
jpslav: See above about aligning the PORO and Rake file/task names.  In any event, this rake file/task name needs to be pluralized to end in "Jobs".

Also, note that any changes to the rake file name need to be reflected in the worker launch configuration https://github.com/openstax/unified-deployment/blob/master/cfn/search/worker.yml#L161

app/services/done_index_job.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def as_json(*)
+    {
+      "type" => @type,
+      "book_version_id" => @book_version_id,
+      "indexing_version" => @indexing_version,
+      "status" => @status
+    }
+  end
```
jpslav: ditto using `{x: y}` symbol hash notation and also the attr_reader methods available.

jpslav (changes requested):

app/services/process_index_jobs.rb
```diff
@@ -0,0 +1,80 @@
...
+      rescue => ex   # TODO review error handling with JP
+        enqueue_done_job(job: job,
+                         status: DoneIndexJob::Status::STATUS_OTHER_ERROR,
+                         message: ex.message)
+      end
+
+      record_job_stat(job)
+    end
+
+    @worker_asg_instance.terminate_instance
```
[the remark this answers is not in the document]
jpslav: May also make sense to move the `increase_desired_capacity` in the `EnqueueIndexJobs` to its rake task.

app/services/search/book_versions/index.rb
steveburkett: it does raise an exception.

spec/models/book_index_state_spec.rb
```diff
@@ -14,12 +14,17 @@
   subject(:book_index_state) { described_class }
 
   describe ".create" do
-    it 'creates the a document row in the dynamo db table' do
+    it 'creates the a document row in the dynamo db table with correct status' do
```
jpslav: `row` => `item` -- so we don't confuse future developers into thinking dynamo is a RDBMS :-)

spec/services/base_index_job_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  class Test
+    def self.foobar
+    end
+  end
+
+  let(:proc) {
+    -> {
+      Test.foobar
+    }
+  }
```
jpslav: A more compact way if you're interested (or feel free to ignore)

```
let(:test) { double("blah") }
let(:proc) { -> { test.foobar } }
...
expect(test).to receive(:foobar).once
```

jpslav (changes requested):
looks good! some comments.  Travis is failing, fyi

app/models/book_index_state.rb
```diff
@@ -7,17 +7,41 @@
...
+    ACTIONS = [
+      ACTION_CREATED = 'enqueued_created',
+      ACTION_DELETED = 'enqueued_deletion'
+    ]
+
+    attr_reader :action, :at
+
+    def initialize(action:, at: DateTime.now)
+      @action = action
+      @at = at
```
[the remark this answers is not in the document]
steveburkett: good catch.  normally, activerecord and dynamoid default to UTC, but here, we're doing the "custom" serialization.

app/models/book_index_state.rb
```diff
@@ -7,17 +7,41 @@
...
-  field :enqueued_time, :datetime, store_as_string: true
-
-  field :updated_at,    :datetime, store_as_string: true
-  field :created_at,    :datetime, store_as_string: true
+  field :status_log, :array,     of: Status
+  field :updated_at, :datetime,  store_as_string: true
+  field :created_at, :datetime,  store_as_string: true
   field :message
 
   STATES                  = [
```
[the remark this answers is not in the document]
steveburkett: STATES  isssssssssssssssssssssssss equal to [  :)

app/models/book_index_state.rb
```diff
@@ -34,12 +58,13 @@ class BookIndexState
...
-      job.indexing_version = indexing_version
-      job.enqueued_time = DateTime.now
-      job.save!
+    new.tap do |book_index_state|
+      book_index_state.state = state
+      book_index_state.book_version_id = book_version_id
+      book_index_state.indexing_version = indexing_version
+      new_status = Status.new(action: Status::ACTION_CREATED)
+      book_index_state.status_log = [ new_status ]
+      book_index_state.save!
```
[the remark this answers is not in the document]
steveburkett: just a personal style thing (https://github.com/rubocop-hq/ruby-style-guide/issues/613).  not a big deal, of course, will change back to the param style.

app/models/todo_jobs_queue.rb
```diff
@@ -8,7 +8,20 @@ def write(index_job)
...
+    parsed_message = JSON.parse(messages.first.body).with_indifferent_access
+    msg_type = parsed_message[:type].constantize
+
+    receipt_handle = messages.first.receipt_handle
+    when_completed_proc = -> {
+      raw_queue.delete_messages({entries: [{id: SecureRandom.uuid,
+                                            receipt_handle: receipt_handle}]})
+    }
+    new_job = msg_type.build_object(body: parsed_message,
+                                    when_completed_proc: when_completed_proc)
```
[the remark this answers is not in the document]
steveburkett: agreed!

app/services/base_index_job.rb
```diff
@@ -0,0 +1,26 @@
+class BaseIndexJob
+  attr_reader :indexing_version,
```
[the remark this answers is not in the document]
steveburkett: this is good.  we could use "index_strategy_instance" to differentiate from "index strategy"

app/services/create_index_job.rb
```diff
@@ -0,0 +1,19 @@
...
+    new(book_version_id: body[:book_version_id],
+        indexing_version: body[:indexing_version],
+        when_completed_proc: when_completed_proc)
+  end
+
+  def call
+    index.recreate
+  end
+
+  def as_json(*)
```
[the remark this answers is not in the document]
steveburkett: @jpslav https://blog.arkency.com/how-to-overwrite-to-json-as-json-in-active-record-models-in-rails/ (the as_json is passed different things in different scenarios.  in this case, i dont think we care...

app/services/done_index_job.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def self.build_object(body:, when_completed_proc:)
+    new(results: body[:results],
+        book_version_id: body[:book_version_id],
+        indexing_version: body[:indexing_version],
+        when_completed_proc: when_completed_proc)
+  end
+
+  def initialize(results:, **opts)
+    super(opts)
```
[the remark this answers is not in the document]
steveburkett: @jpslav but JP...it's so ruby-cool to slurp up all the extra args into a hash.... :)

app/services/search/book_versions/i1/indexing_strategy.rb
```diff
@@ -27,7 +27,7 @@ def index_metadata
     end
 
     def index(book:, index_name:)
-      Rails.logger.info("OpenSearch: indexing #{index_name} with #{book.root_book_part.pages.count} pages")
+      Rails.logger.info("I1::IndexingStrategy: Creating index #{index_name} with #{book.root_book_part.pages.count} pages")
```
jpslav: Just noticed that the `pages` method isn't memoized.  Can you make that change in cnx-ruby?

jpslav (changes requested):
Looks great!

jpslav (approved):

PR #25, opened 2019-05-20
Formalize search API

jpslav:
Transition from the prototype search query endpoint to the "real" endpoint.

* Prototype endpoints have been moved to a `PrototypeController`
* Added Swagger documentation.  Swagger JSON is available at `/api/v0/swagger`.  
* Added `openstax_swagger` gem usage (which is an extraction of code used for `a15k/interactions-api`).  This lets us document the API with Swagger and also lets us generate PORO bindings for our API interfaces (lets us parse and export JSON by just dealing with POROs)
* Wrote the "real" search endpoint with a request spec for it.  While our main goal is to search one book at a time, the search endpoint takes an array of books to make it possible to search across books should we want to in the future.
* Added "search strategies" to complement "index strategies".  Whereas index strategies control how a book is indexed into Elasticsearch, search strategies control how we query that content out of Elasticsearch.  Implemented the "S1" search strategy to match the search requirements from UX.  See the [simple_query_string](https://www.elastic.co/guide/en/elasticsearch/reference/6.3/query-dsl-simple-query-string-query.html) reference page for details on the options we are using.
* Currently, the index strategy and search strategy are provided as inputs to the search endpoint.  At some point in the future we may not do that, but for now it gives us flexibility as we experiment with search.
* Added a rake task to generate the files with the Swagger JSON (see the README)

(rest of todos moved to another separate issue)


app/controllers/api/v0/search_controller.rb
```diff
@@ -1,47 +1,64 @@
...
+      end
+      parameter do
+        key :name, :book
+        key :in, :query
+        key :description, 'UUID@version of the book to search'
+        key :required, true
+        key :type, :string
+      end
+      parameter do
+        key :name, :index_strategy
```
TomWoodward: didn't we discuss this being fixed per api version and the client not needing to care about it?

app/controllers/api/v0/search_controller.rb
```diff
@@ -1,47 +1,64 @@
...
+        key :in, :query
+        key :description, 'name of the index strategy to use when searching'
+        key :required, true
+        key :type, :string
+      end
+      parameter do
+        key :name, :search_strategy
```
TomWoodward: same for this?

app/controllers/api/v0/search_controller.rb
```diff
@@ -1,47 +1,64 @@
...
-       "highlight": {
-          "fields": {
-            "content": {}
-          }
-        }
-    }
-
-    response = OpenSearch::ElasticsearchClient.instance.search body: query.to_json
-    render json: response
+  swagger_path '/search' do
```
TomWoodward: so this path is going to end up looking something like `openstax.org/search/api/v0/search` right?

TomWoodward (commented):
i in general don't love committing generated code, but i'll defer to you ruby guys

app/controllers/api/v0/search_controller.rb
```diff
@@ -1,47 +1,64 @@
...
+      key :operationId, 'search'
+      parameter do
+        key :name, :q
+        key :in, :query
+        key :description, 'text of the search query'
+        key :required, true
+        key :type, :string
+      end
+      parameter do
+        key :name, :book
```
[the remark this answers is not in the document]
jpslav: You mean searching across multiple books?  We don't have that use case envisioned do we?

We could make the parameter be an array and freak out if there's more than one entry for the time being.

app/controllers/api/v0/search_controller.rb
jpslav: Yes... `openstax.org/open-search/api/v0/search`.  The `/api/v0` part is defined as an API base somewhere else... but this is a good point that we don't have the `open-search` prefix there.  We could add it and just make all Swagger use assume access via CloudFront (and that prefix) or we could have two Swagger JSONs, one with and one without the prefix.

app/controllers/api/v0/search_controller.rb
jpslav: Let's maybe park this.  I do recall some conversation related to this, but there are reasons to not have a different API version for every index strategy / search strategy combination.  As it is currently...

* The API version controls the format of the requests and the top-level format of the response
* The index strategy controls what is indexed and how it is indexed (e.g. page elements, indexing options, etc)
* The search strategy controls how we search the index (fuzzy vs not fuzzy, boosting fields, full matches vs OR matches, etc)

The API is returning raw ES results plus some top-level metadata.  The top-level metadata is fixed for each API version.  The raw ES results depend on the combination of strategies.

One API version, defining the request format and top-level response format, could work with several combinations of strategies.  

There likely won't be a zillion strategies, and the Rex FE would change slowly over time to use different ones (e.g. the `i1` indexing strategy could be used for many months).  

The `search_strategy` parameter would give us a way to A/B test different search queries (with the A/B machinery being external to search itself).

app/controllers/api/v0/search_controller.rb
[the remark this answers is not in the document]
TomWoodward: i don't think UI has any plans to implement, but if it was easy to do on the back we could just do it, if not, whatever

app/controllers/api/v0/search_controller.rb
TomWoodward: i was thinking that indexing/search strategy changes that did not influence the structure of the responses could be done within one api version, and any other change would create an api version. exposing 3 versions of things for the client to manage seems like a lot, and i was actually thinking it would be nicer to handle a/b testing of search stuff within search, rather than forcing the client to deal with all that

app/controllers/api/v0/search_controller.rb
TomWoodward: we can just use the prefix all the time right? no reason to support queries without the prefix like accounts does

app/controllers/api/v0/search_controller.rb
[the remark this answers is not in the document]
jpslav: I updated this to take a list of books, and it *should* work, tho it isn't tested :-o

app/controllers/api/v0/search_controller.rb
jpslav: After a discussion with Tom, we decided it is ok for the time being to leave the query parameters that choose index strategy and search strategy.

app/controllers/api/v0/search_controller.rb
jpslav: I updated to use the `/open-search` prefix in the swagger json

config/initializers/aws.rb
```diff
@@ -12,8 +12,20 @@ def aws_creds
   end
 end
 
-Aws.config.update(
-  {
+def use_aws?
+  !Rails.env.development? ||
+  ENV['USE_AWS'] ||
```
steveburkett: not sure if USE_AWS is needed....

config/initializers/openstax_swagger.rb
```diff
@@ -0,0 +1,10 @@
+OpenStax::Swagger.configure do |config|
+  config.json_proc = -> (api_major_version) {
+    Swagger::Blocks.build_root_json(
+      "Api::V#{api_major_version}::Swagger::DocsController::SWAGGERED_CLASSES".constantize
+    )
+  }
+end
+
```
steveburkett: extra lines

spec/requests/api/v0/search_requests_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      json = json_response
+      expect(json[:overall_took]).not_to be_nil
+      expect(json[:hits][:total]).to eq 1
+      expect(json[:hits][:hits][0][:_source]).to include(
+        page_id: "2c60e072-7665-49b9-a2c9-2736b72b533c@8",
+        element_type: "paragraph",
+        page_position: 3
+      )
+      expect(json[:hits][:hits][0][:highlight][:visible_content][0]).to start_with "<em>Recall</em>"
```
steveburkett: uggggg

app/controllers/api/v0/prototype_controller.rb
```diff
@@ -0,0 +1,47 @@
+class Api::V0::PrototypeController < Api::V0::BaseController
+
+  def search
+    # TODO generalize
```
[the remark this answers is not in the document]
jpslav: This class will be deleted in a couple of weeks

app/controllers/api/v0/prototype_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+            "content": {}
+          }
+        }
+    }
+
+    response = OpenSearch::ElasticsearchClient.instance.search body: query.to_json
+    render json: response
+  end
+
+  def temp_build_index
```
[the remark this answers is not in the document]
jpslav: This class will be deleted in a couple of weeks

Gemfile
```diff
@@ -45,7 +45,7 @@ gem 'dotenv-rails'
 
 gem "openstax_path_prefixer", github: "openstax/path_prefixer", ref: "b6d8f45d8"
 
-gem "openstax_aws", github: "openstax/aws-ruby", ref: '1554be9d36df9f8129'
+gem "openstax_aws", github: "openstax/aws-ruby", ref: '1554be9d36'
```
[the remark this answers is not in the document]
jpslav: I think 7 digits or more is good.  More digits makes (extremely unlikely) collisions less probable.

PR #26, opened 2019-05-30
use credentials provider instead of credentials to get refreshes

jpslav:
Previously we were passing in static credentials for making signed AWS requests to Elasticsearch.  This PR instead provides an object that provides credentials on every request to Elasticsearch.  That retrieval of credentials is smart and refreshes the returned credentials before they expire.


steveburkett (approved):
looks good afaict.

TomWoodward (approved):
/shrug

PR #27, opened 2019-05-30
Monitoring indexing jobs basic objects

steveburkett:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/199

Add a monitoring rake task to process the Done queue by setting correct book index state when done. Also, handle sending errors to Sentry and keeping an eye on the auto scaling group desired capacity.


app/services/base_index_job.rb
```diff
@@ -3,24 +3,46 @@ class BaseIndexJob
...
+    book_index = find_associated_book_index
+    book_index.destroy!
   end
 
-  def when_completed
-    @when_completed_proc.try(:call)
+  def metadata
+    internal_data = JSON.parse(to_json)
+    book_index = find_associated_book_index
+    internal_data.merge(JSON.parse((book_index || {}).to_json))
```
jpslav: Minor but putting `to_hash` methods on `BaseIndexJob` and `BookIndexState` would clean this to

```
def metadata
  to_hash.merge(find_associated_book_index.to_hash)
end
```

app/services/base_index_job.rb
```diff
@@ -3,24 +3,46 @@ class BaseIndexJob
...
   end
 
-  protected
+  def find_associated_book_index
```
jpslav: Please add `_state` to the end so it is `find_associated_book_index_state` (for consistency)

app/services/base_index_job.rb
```diff
@@ -3,24 +3,46 @@ class BaseIndexJob
...
+  def remove_associated_book_index
+    book_index = find_associated_book_index
+    book_index.destroy!
   end
 
-  def when_completed
-    @when_completed_proc.try(:call)
+  def metadata
+    internal_data = JSON.parse(to_json)
+    book_index = find_associated_book_index
```
jpslav: per below, make the var `book_index_state`.

You could also consider just making the method be `book_index_state`

app/services/create_index_job.rb
```diff
@@ -1,14 +1,13 @@
...
-        when_completed_proc: when_completed_proc)
-  end
-
   def call
     index.recreate
   end
 
+  def cleanup_when_done
+    book_index = find_associated_book_index
+    book_index.mark_created
```
jpslav: Liking changing `find_associated_book_index` to `book_index_state`, which makes this method be `book_index_state.mark_created`

app/services/delete_index_job.rb
```diff
@@ -1,14 +1,12 @@
...
-        indexing_strategy_name: body[:indexing_strategy_name],
-        when_completed_proc: when_completed_proc)
-  end
-
   def call
     index.delete
   end
 
+  def cleanup_when_done
+    remove_associated_book_index
```
jpslav: similar to above... could be `book_index_state.destroy!` ... tho I do see your other uses for this `remove_...` convenience method

app/services/done_index_job.rb
```diff
@@ -1,57 +1,64 @@
...
-    super(book_version_id: book_version_id,
-          indexing_strategy_name: indexing_strategy_name,
-          when_completed_proc: when_completed_proc)
+  def initialize(status: STATUS_SUCCESSFUL,
+                 es_stats: nil,
+                 time_took: nil,
+                 ran_job: nil,
+                 message: nil,
+                 cleanup_after_call: nil)
+    super(cleanup_after_call: cleanup_after_call)
```
jpslav: if you pass `book_version_id` and `index_strategy_name` here with the values from `ran_job` then this done job's `attr_reader` for those fields will return something and you won't have to override `find_associated_book_index` below.

app/services/base_index_job.rb
```diff
@@ -3,24 +3,46 @@ class BaseIndexJob
               :book_version_id,
```
jpslav: Did we switch to `index_strategy` being a string and `index_strategy_instance` being the object?  If so, should we go through and get all the `index_strategy_name`s out?

app/services/done_index_job.rb
```diff
@@ -1,57 +1,64 @@
...
-      {
-        status: status,
-        es_status: es_stats,
-        time_took: time_took,
-        message: message
-      }
-    end
+  def find_associated_book_index
+    BookIndexState.where(book_version_id: ran_job.book_version_id,
+                         indexing_strategy_name: ran_job.indexing_strategy_name).first
```
jpslav: See above, this override potentially not needed

app/services/monitor_index_job.rb
```diff
@@ -0,0 +1,59 @@
+# Monitor the Done jobs from the SQS queue
+class MonitorIndexJob
```
jpslav: pluralize this class name so it doesn't look like a job class.

app/services/monitor_index_job.rb
```diff
@@ -0,0 +1,59 @@
...
+  def initialize
+    @todo_jobs_queue = TodoJobsQueue.new
+    @done_jobs_queue = DoneJobsQueue.new
+    @dead_queue = DeadLetterJobsQueue.new
+    @processed_from_done = 0
+    @aws_asg = OpenStax::Aws::AutoScalingGroup.new(
```
jpslav: since there are 3 ASGs, recommend naming this `@worker_asg` for clarity

app/services/monitor_index_job.rb
```diff
@@ -0,0 +1,59 @@
...
+      done_job = @done_jobs_queue.read
+      break if done_job.nil?
+
+      Rails.logger.info("MonitorIndexJob: job #{done_job.class.to_s} #{done_job.to_json} started...")
+
+      done_job.call
+      done_job.cleanup_after_call
+
+      @processed_from_done +=1
+    end
```
jpslav: Put this loop into a private method like the others, e.g. `process_done_jobs`?

app/services/monitor_index_job.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def call
+    loop do
+      done_job = @done_jobs_queue.read
+      break if done_job.nil?
+
+      Rails.logger.info("MonitorIndexJob: job #{done_job.class.to_s} #{done_job.to_json} started...")
+
+      done_job.call
+      done_job.cleanup_after_call
```
jpslav: This pattern of calling `call` and then `cleanup_after_call` makes me wonder if the `cleanup_after_call` call should be automatic?  The `BaseIndexJob` has the cleanup_after_call proc in its possession, so we could do...

```ruby
class BaseIndexJob
  def call
    _call
    @cleanup_after_call.try(:call)
  end

  def _call
    raise "Implement _call in the child class!"
  end
end
```

Would this work?

app/services/monitor_index_job.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  private
+
+  def notify_sentry_for_dead_letter_queue
+    loop do
+      dead_job = @dead_queue.read
+      break if dead_job.nil?
+
+      Raven.capture_message("Job Found in Dead Letter Queue", :extra => dead_job.metadata)
```
jpslav: The `DoneIndexJob`'s `call` method contains the Raven call.  Should the dead letter job have this raven call in it, similarly?  If not, then we should consider pulling out the Raven call from the done index job for consistency.

app/services/monitor_index_job.rb
```diff
@@ -0,0 +1,59 @@
...
+
```
jpslav: Keep counter for stats here too?

app/services/work_index_job.rb
```diff
@@ -35,24 +35,24 @@ def call
...
-                       status: DoneIndexJob::Results::STATUS_OTHER_ERROR,
+                       status: DoneIndexJob::STATUS_OTHER_ERROR,
                        message: ex.message)
     end
 
     job_stats
   end
 
-  private
+  # private
```
jpslav: Uncomment?

config/initializers/raven.rb
```diff
@@ -0,0 +1,7 @@
+if Rails.env.production? || ENV['FORCE_ENABLE_SENTRY'] == 'true'
```
jpslav: Please add note about `FORCE_ENABLE_SENTRY` to the README.  README might need a new configuration section.

lib/tasks/monitor_index_jobs.rake
```diff
@@ -0,0 +1,9 @@
+desc <<-DESC.strip_heredoc
+  Monitors the done sqs queue.
+DESC
+task monitor_index_jobs: :environment do
+  Rails.logger.info { "Starting monitor_index_jobs..." }
+
+  monitor_index_job = MonitorIndexJob.new
+  monitor_index_job.call
```
jpslav: Log stats?

jpslav (changes requested):
Looks nice

app/services/monitor_index_job.rb
steveburkett: i had left it this way in case one wanted to not call cleanup_after_call, keeping the message on the queue for some reason.  but if it's always to be removed, then yea, maybe.

app/services/base_index_job.rb
steveburkett: @jpslav i thought we settled on the more semantic "index_strategy_name" (representing "I1").  just like search_strategy_name would be "S1".   and index_strategy and search_strategy would be the objects.

app/services/base_index_job.rb
steveburkett: renamed the method inspect

app/services/monitor_index_job.rb
steveburkett: @jpslav perhaps this is a candidate for call! and call?  where call! deletes the message after the #call?

jpslav:
When would we want to not delete the message after processing?

app/models/book_index_state.rb
```diff
@@ -75,14 +75,24 @@ def initialize(*args)
...
     save!
   end
 
   def deleting?
-    [STATE_DELETED, STATE_DELETE_PENDING].include?(self.state)
+    [STATE_DELETE_PENDING].include?(self.state)
+  end
+
+  def to_hash
+    JSON.parse(to_json)
```
jpslav: Add `.with_indifferent_access` to match the other `to_hash`?

app/services/base_index_job.rb
```diff
@@ -3,24 +3,59 @@ class BaseIndexJob
...
-    @book_uuid, @book_version = @book_version_id.split('@')
+  def to_hash
+    JSON.parse(to_json).with_indifferent_access
   end
 
-  def when_completed
-    @when_completed_proc.try(:call)
+  def remove_associated_book_index_state
+    book_index_state = find_associated_book_index_state
+    book_index_state.destroy!
```
jpslav: what if `book_index_state` is nil?

app/services/base_index_job.rb
```diff
@@ -3,24 +3,59 @@ class BaseIndexJob
...
   end
 
-  protected
+  def inspect
+    to_hash.merge(find_associated_book_index_state.to_hash)
```
jpslav: love it

app/services/create_index_job.rb
```diff
@@ -1,12 +1,7 @@
...
-    new(book_version_id: body[:book_version_id],
-        indexing_strategy_name: body[:indexing_strategy_name],
-        when_completed_proc: when_completed_proc)
-  end
-
-  def call
-    index.recreate
+  def cleanup_when_done
+    book_index_state = find_associated_book_index_state
+    book_index_state.mark_created
```
jpslav: what if nil? (shouldn't be, but...)

app/services/monitor_index_jobs.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def stats
+    {
+      processed_from_done: @processed_from_done,
+      desired_capacity_reset_by: @desired_capacity_reset_by,
+    }
+  end
+
+  private
```
jpslav: extra unneeded `private`?

app/services/monitor_index_jobs.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def notify_sentry_for_dead_letter_queue
+    loop do
+      dead_job = @dead_queue.read
+      break if dead_job.nil?
+
+      Raven.capture_message("Job Found in Dead Letter Queue", :extra => dead_job.inspect)
```
jpslav: Should probably also have a little log message that we are sending to sentry (unless sentry does that for us) to be consistent with the other private method logging.  Don't need/want to log the `.inspect` content, just say that we are sending something to sentry, maybe include the index name or something identifiable.

jpslav (changes requested):
Looks great!

jpslav (approved):

PR #28, opened 2019-06-06
Add element id to search

steveburkett:
Add element ID to search results. 

To do this, add the element ID to indexing and the swagger response schema.   Added the ID as a method to the figure and paragraph objects in cnx-ruby as well.   

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/298


app/services/search/book_versions/i1/indexing_strategy.rb
```diff
@@ -40,9 +40,16 @@ def index_page(page:, index_name:)
...
-                                           body: document.body)
+        if document.ok_to_index?
+          OpenSearch::ElasticsearchClient.instance.index(index: index_name,
+                                                         type: document.type,
+                                                         body: document.body)
+        else
+          message = "Search::BookVersions::I1::IndexingStrategy: unable to index "/
+            "Page id #{element.page_id}, #{element.element_type} due to no element ID"
+          Raven.capture_message(message, :extra => element.to_json)
+          Rails.logger.warn(message)
```
jpslav: The error message here doesn't obviously equate to `!document.ok_to_index?` ... the `if` condition is generic but the message is specific.

What about making `document.body` raise an exception if the `element_id` is blank, and then here just have `begin;rescue;end` that catches the exception (which can contain a specific message about the element ID being blank) and sends it to Sentry?

jpslav (changes requested):

app/services/search/book_versions/i1/indexing_strategy.rb
steveburkett: @jpslav very true.   I agree.  This is the classic case of a good use of an exception over writing defensive code.

jpslav (approved):

PR #29, opened 2019-06-06
fix install_secrets to handle multi-level secrets namespaces

jpslav:
When secrets namespaces are multi-level, e.g. "search/api", there are more parts of Parameter Store keys to ignore when getting the actual secrets out of the Parameter Store.  This PR accounts for such multi-level namespaces.

### DIFF
--- a/lib/tasks/install_secrets.rake
+++ b/lib/tasks/install_secrets.rake
@@ -30,13 +30,17 @@ task :install_secrets, [] do
 
   secrets = {}
 
+  # When we get parameters from the store, we want to ignore the env_name
+  # and the parameter namespace.  Calculate how many levels there are here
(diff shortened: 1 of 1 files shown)


lib/tasks/install_secrets.rake
```diff
@@ -30,13 +30,17 @@ task :install_secrets, [] do
 
   secrets = {}
 
+  # When we get parameters from the store, we want to ignore the env_name
+  # and the parameter namespace.  Calculate how many levels there are here
+  num_ignored_key_levels = [env_name, namespace].join('/').split('/').count
```
steveburkett: should these 2 "utils" be put into a utility method so it can be unit tested (and potentially reused?)

steveburkett (approved):
just one comment that could go either way

PR #31, opened 2019-06-08
make workers keep looking for termination

jpslav:
If an exception happens when an instance tries to terminate, it will no longer be able to terminate (and it'll just loiter until its failsafe `shutdown` call kicks in 3h later).  This PR lets it continue trying to terminate.  

also, send exceptions off to Sentry as soon as we get them instead of waiting for them to be processed in the Done job.  We were not getting enough exception information out.

### DIFF
--- a/app/services/work_index_jobs.rb
+++ b/app/services/work_index_jobs.rb
@@ -38,10 +38,12 @@ class WorkIndexJobs
                        status: DoneIndexJob::STATUS_SUCCESSFUL,
                        es_stats: es_stats,
                        time_took: time_took)
-    rescue InvalidIndexingStrategy
+    rescue InvalidIndexingStrategy => ex
+      Raven.capture_exception(ex)
       enqueue_done_job(job: job,
                        status: DoneIndexJob::STATUS_INVALID_INDEXING_STRATEGY)
     rescue => ex
+      Raven.capture_exception(ex)
       enqueue_done_job(job: job,
                        status: DoneIndexJob::STATUS_OTHER_ERROR,
                        message: ex.message)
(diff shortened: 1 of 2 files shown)


app/services/work_index_jobs.rb
```diff
@@ -38,10 +38,12 @@ def call
                        status: DoneIndexJob::STATUS_SUCCESSFUL,
                        es_stats: es_stats,
                        time_took: time_took)
-    rescue InvalidIndexingStrategy
+    rescue InvalidIndexingStrategy => ex
+      Raven.capture_exception(ex)
```
steveburkett: so here, with this, we'll end up with 2 Sentry exceptions:  one from here, and then one later, when the done job is processed w/ a INVALID_INDEXING_STRATEGY.   Is this ok?

lib/tasks/work_index_jobs.rake
```diff
@@ -2,38 +2,39 @@ desc <<-DESC.strip_heredoc
...
+      elsif work_index_job.definitely_out_of_work?
+        Rails.logger.info { "work_index_jobs: out of work, terminating"}
+        instance.terminate(should_decrement_desired_capacity: true, continue_hook_name: "TerminationHook")
+        break
+      else
+        Rails.logger.info { "work_index_jobs: starting #call"}
+        stats = work_index_job.call # reads from queue, works the job, writes to done queue
+        Rails.logger.info { "work_index_jobs: #call finished #{stats.to_s}" }
+      end
+    rescue Aws::AutoScaling::Errors::ScalingActivityInProgress => ee
```
steveburkett: why ee?  usually, the general pattern for exceptions is 'ex' or 'exception' or even 'e'.  why a new one ee? :)

jpslav: Two Es are better than one.  Duh!

jpslav: I agree `e` was the original common way, but one letter variables are the root of all evil, so at some point in my distant past I doubled it to `ee`.

app/services/work_index_jobs.rb
jpslav: Yes

PR #34, opened 2019-06-13
rescue errors when processing dead letter jobs

jpslav:
* Rescue errors when processing dead letter jobs (just in case)
* Add a `logger_prefix` utility to prefix log messages with some string within a class

### DIFF
--- a/app/services/monitor_index_jobs.rb
+++ b/app/services/monitor_index_jobs.rb
@@ -1,10 +1,14 @@
 # Monitor the Done jobs from the SQS queue
 class MonitorIndexJobs
+
+  prefix_logger "MonitorIndexJobs"
+
   def initialize
     @todo_jobs_queue     = TodoJobsQueue.new
     @done_jobs_queue     = DoneJobsQueue.new
     @dead_queue          = DeadLetterJobsQueue.new
     @processed_from_done = 0
+    @processed_from_dead = 0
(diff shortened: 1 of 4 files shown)


app/services/monitor_index_jobs.rb
```diff
@@ -26,10 +30,17 @@ def notify_sentry_for_dead_letter_queue
...
+        log_info("Sending job #{dead_job.class.to_s} to dead letter queue")
+        Raven.capture_message("Job Found in Dead Letter Queue", :extra => dead_job.inspect)
+
+        dead_job.cleanup_after_call   #delete the message so we dont keep sending to Sentry
 
-      dead_job.cleanup_after_call   #delete the message so we dont keep sending to Sentry
+        @processed_from_dead += 1
+      rescue => ex
+        Raven.capture_exception(ex)
+        log_error("dead job error #{ex.message} process #{dead_job.inspect}")
```
nathanstitt: somehow I'd never encountered `log_error` and `_info` - TIL they're part of action_controller which I'm guessing is mixed in to here?

nathanstitt (approved):

app/services/monitor_index_jobs.rb
jpslav: This is from the `prefix_logger.rb` in this PR

PR #35, opened 2019-07-03
search api part 2 of 2

steveburkett:
Finish the search API story.   

- [x] Add tests for search factory and strategy
- [x] Reorganize the services code.
- [x] Re-record the cassettes.
- [x] Update max search results 
- [x] Update elastic search num of highlight fragments to return

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/319


app/controllers/api/v0/search_controller.rb
```diff
@@ -64,7 +64,7 @@ class Api::V0::SearchController < Api::V0::BaseController
   def search
     started_at = Time.now
 
-    search_strategy_instance = Search::BookVersions::SearchStrategies::Factory.build(
+    search_strategy_instance = Books::SearchStrategies::S1::Factory.build(
```
jpslav: The factory was responsible for building the right instance of the search strategy, so I don’t think it can be inside the `S1` namespace?  (because it can create an `S2` strategy if that is what is passed in in the `search_strategy` argument)

app/controllers/api/v0/search_controller.rb
```diff
@@ -1,8 +1,8 @@
 class Api::V0::SearchController < Api::V0::BaseController
 
   EXCEPTIONS_422 = [
-    Search::BookVersions::SearchStrategies::UnknownSearchStrategy,
-    Search::BookVersions::SearchStrategies::IncompatibleStrategies
+    Books::SearchStrategies::S1::UnknownSearchStrategy,
```
jpslav: These should not be scoped under `S1`.  They are common across strategies of which `S1` is one.

app/services/books/index.rb
```diff
@@ -1,12 +1,12 @@
-module Search::BookVersions
-  # Search::BookVersions::Index is the main interface into indexing a book version.
+module Books
+  # Books::Index is the main interface into indexing a book version.
```
jpslav: recommend removing “version” from the end of this comment since we took it out of the namespace name.

app/services/books/index.rb
```diff
@@ -1,12 +1,12 @@
...
   #
   # It uses a IndexingStrategy that defines what is indexed including the
   # index's inspect.
   #
-  # This class perf gueorms the "crud" actions on a book's index.
+  # This class performs the "crud" actions on a book's index.
```
jpslav: gueorms!!

app/services/books/search_strategies/s1/factory.rb
```diff
@@ -1,8 +1,8 @@
-module Search::BookVersions::SearchStrategies
+module Books::SearchStrategies::S1
   class Factory
 
     STRATEGY_CLASSES = [
-      S1
+      Strategy
```
jpslav: When this factory moves back up then this because `STRATEGY_CLASSES = [S1::Strategy]`

app/services/work_index_jobs.rb
```diff
@@ -40,10 +40,12 @@ def call
                        time_took: time_took)
     rescue InvalidIndexingStrategy => ex
       Raven.capture_exception(ex)
+      Rails.logger.error(ex)
       enqueue_done_job(job: job,
                        status: DoneIndexJob::STATUS_INVALID_INDEXING_STRATEGY)
     rescue => ex
       Raven.capture_exception(ex)
+      Rails.logger.error(ex)
```
jpslav: How do these look in the production logs?

app/services/base_index_job.rb
```diff
@@ -21,6 +21,7 @@ def initialize(book_version_id: nil,
 
   def call
     _call
+  ensure #it should always remove the job from the queue
```
jpslav: Can you exercise this change in a spec, even with some stubbing in a dummy job where the `cleanup_after_call` proc is a spy that you can assert receives `:call`?

jpslav (changes requested):
Like the reorg and changes!  Made only a few comments.

jpslav (approved):

PR #36, opened 2019-07-08
make hits required

TomWoodward:


steveburkett:
we should prob. add a top level controller test for this too eh?  one that tests for an empty response, meeting the requirement of hits being required?

jpslav:
@steveburkett do you mean testing that elasticsearch returns an empty hits array?

TomWoodward:
@steveburkett @jpslav not sure if that is necessary since its really testing elasticsearch

jpslav:
@TomWoodward steve brought up a good point that we should be making sure that what we give back to our clients matches the API spec we are publishing, so I made another card for steve to implement some better verification that we are getting from ES and putting into our API output is actually what we publish.

PR #37, opened 2019-07-10
bogus rex releases

steveburkett:
Change the work index jobs service to handle cxn book 404s by creating a new done status for this error and processing different in the monitor service.   When a http 404 error is received, keep the book index state around so that it can be fixed (not re-indexed or deleted).  

Also, add pre_ and post_tags to change the highlighting markup returned by elasticsearch. 

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/318

### DIFF
--- a/app/models/book_index_state.rb
+++ b/app/models/book_index_state.rb
@@ -11,6 +11,7 @@ class BookIndexState
     ACTIONS = [
       ACTION_CREATE  = 'enqueued_create',
       ACTION_CREATED = 'index_created',
+      ACTION_ERROR   = 'index_error',
       ACTION_DELETE_PENDING = "index_delete_pending"
     ]
 
@@ -48,6 +49,7 @@ class BookIndexState
   STATES = [
     STATE_CREATE_PENDING = "create pending",
     STATE_DELETE_PENDING = "delete pending",
+    STATE_HTTP_ERROR = "http error",
     STATE_CREATED = "created"
   ]
   VALID_INDEXING_STRATEGY_NAMES = %w(I1)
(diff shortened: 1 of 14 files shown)


app/models/book_index_state.rb
```diff
@@ -81,6 +83,12 @@ def mark_created
     save!
   end
 
+  def mark_as_http_error
+    self.state = STATE_HTTP_ERROR
+    self.status_log << StatusLog.new(action: StatusLog::ACTION_ERROR)
```
jpslav: when we look in dynamo we won't see that it was an http error (will say "index_error").  should it say "http_error"?  or add "http_error" and then add "index_error"?

app/services/done_index_job.rb
```diff
@@ -63,8 +68,14 @@ def _call
     if successful?
       ran_job.cleanup_when_done
     else
-      Raven.capture_message("Job in Error Found in Done Queue", :extra => inspect)
-      ran_job.remove_associated_book_index_state
+      if http_error?
+        Raven.capture_message("Openstax http error", :extra => inspect)
```
jpslav: Can remove "Openstax".

What does a second Raven capture get for us here? (with the other one in the worker)

spec/models/book_index_state_spec.rb
```diff
@@ -2,6 +2,7 @@
 require 'vcr_helper'
 
 amazon_api_header_matcher = lambda do |request_1, request_2|
+  # puts "request1 x-amz-target #{request_1.headers["X-Amz-Target"]}, request2 x-amz-target #{request_2.headers["X-Amz-Target"]}"
```
jpslav: puts

jpslav (changes requested):

PR #39, opened 2019-07-23
Fix Elasticsearch exceptions

steveburkett:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/392

### DIFF
--- a/app/services/books/index.rb
+++ b/app/services/books/index.rb
@@ -22,25 +22,31 @@ module Books
       @indexing_strategy = indexing_strategy.new
     end
 
-    def create
-      OsElasticsearchClient.instance.indices.create(index: name, body: @indexing_strategy.index_metadata)
+    def create(with_wait: false)
+      Rails.logger.debug("Books::Index#create #{name} called")
+      OsElasticsearchClient.instance.indices.create(index: name,
+                                                    body: @indexing_strategy.index_metadata)
+      wait_until(:exists?) if with_wait
     end
 
     # This method populates the index with pages from the book
     def populate
+      Rails.logger.debug("Books::Index#populate #{name} called")
       @indexing_strategy.index(book: book, index_name: name)
 
       index_stats
     end
 
     def recreate
-      delete rescue Elasticsearch::Transport::Transport::Errors::NotFound
-      create
+      delete(with_wait: true) rescue Elasticsearch::Transport::Transport::Errors::NotFound
+      create(with_wait: true)
       populate
     end
 
-    def delete
+    def delete(with_wait: false)
+      Rails.logger.debug("Books::Index#delete #{name} called")
       OsElasticsearchClient.instance.indices.delete(index: name)
+      wait_until(:not_exists?) if with_wait
     end
 
     def name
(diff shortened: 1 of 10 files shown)


app/services/books/index.rb
```diff
@@ -52,8 +58,22 @@ def exists?
...
+    end
+
     private
 
+    def wait_until(this_happens)
+      tries = 1
+      until self.send(this_happens) || tries > 5  do
+        Rails.logger.debug("Waiting for #{name} 1 sec for #{name} to #{this_happens.to_s}, tries: #{tries}")
+        sleep(2)
+        tries += 1
```
jpslav: Is 10 seconds a long enough timeout?  And if we do hit the timeout should we raise an exception?  Otherwise it'll just explode later (at least our code will)

jpslav (changes requested):

jpslav (changes requested):
The expect `Books::Index` spec should exercise `with_wait: true` so that the waiting code has some spec mileage.

app/services/books/index.rb
steveburkett: yeah, let's raise it to 1 minute and raise an exception (a minute would justify an exception)

app/services/books/indexing_strategies/i1/strategy.rb
```diff
@@ -31,6 +31,12 @@ def index(book:, index_name:)
       book.root_book_part.pages.each {|page| index_page(page: page, index_name: index_name) }
     end
 
+    def total_number_of_elements_to_index(book:)
+      num_total = 0
+      book.root_book_part.pages.each{|page| num_total += page.elements(element_classes: DESIRED_ELEMENTS_TO_DOCUMENTS.keys).count }
+      num_total
```
jpslav: This works, can also do `.map{...}.sum`.

Are we worried at all about parsing each page twice? Once here, once during indexing.

Can you also check me to see that pages and root book part is memoized (so we aren't actually downloading the book twice)

app/services/books/index.rb
```diff
@@ -52,8 +63,27 @@ def exists?
...
+      tries = 1
+      until self.send(this_happens) || tries > WAIT_UNTIL_MAX_TRIES  do
+        Rails.logger.debug("Books::Index. Waiting #{WAIT_UNTIL_TRIES_INTERVAL_SEC} secs for #{name} to #{this_happens.to_s}, num_tries so far: #{tries}")
+        sleep(WAIT_UNTIL_TRIES_INTERVAL_SEC)
+        tries += 1
+      end
+
+      if tries >= WAIT_UNTIL_MAX_TRIES
+        raise IndexResourceNotReadyError.new("Books::Index. #{name}:#{this_happens.to_s} failed after #{tries} tries.")
+      end
```
jpslav: This could be at the end of the `until` loop above (changing `>=` to `>` probably which would let you remove the `|| tries > WAIT_UNTIL_MAX_TRIES` there.

spec/cassettes/Books_Index/_name/derives_the_index_name.yml
```diff
@@ -18,6 +18,11 @@ http_interactions:
       code: 404
       message: Not Found
     headers:
+      Warning:
+      - 299 Elasticsearch-6.7.2-56c6e48 "[types removal] The parameter include_type_name
+        should be explicitly specified in get indices requests to prepare for 7.0.
+        In 7.0 include_type_name will default to 'false', which means responses will
+        omit the type name in mapping definitions."
```
jpslav: Good to know for the future :-)

spec/cassettes/Books_Index/_delete/deletes_the_index.yml
```diff
@@ -80,7 +95,39 @@ http_interactions:
...
+    headers:
+      User-Agent:
+      - Faraday v0.15.3
+      Content-Type:
+      - application/json
+      Expect:
+      - ''
+  response:
+    status:
+      code: 200
```
jpslav: Nice to see the check for deletion failing.

jpslav (approved):

app/services/books/indexing_strategies/i1/strategy.rb
steveburkett: def pages
      @pages ||= parts.map do |part|
        part.is_a?(Page) ? part : part.pages
      end.flatten
    end

PR #40, opened 2019-07-30
Eliminate math from indexing

steveburkett:
Replace <math> inside elements with \u2026.

Remove all nodes that contain math.  Replace those removed nodes with a text node containing the unicode ellipses (\u2026).

Add a spec to search spec that verifies mathml replacement and nonexistence in indexing.


app/services/books/indexing_strategies/i1/paragraph_document.rb
```diff
@@ -12,6 +12,8 @@ def self.element_type
     end
 
     def visible_content
+      replace_mathml_nodes
```
jpslav: This is also called in `PageElementDocument::initialize` - duplicate?

jpslav (changes requested):

jpslav (approved):

PR #41, opened 2019-08-01
309: clean up logs

steveburkett:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/309

Several improvements to logging
1. Default production logging to :info
1. Configure logging to show timestamp and pid
2. Return 404 from search API (instead of 500)
3. Serialize job information to Sentry for all errors

_Notes:
We can silence additional Raven or Dynamo (:info or higher) IF we want to use a different logger than the rails logger (logging to a different file) so we can set the level on it to higher than the default rails logging._ 

Image of a 404 & a successful 200 search (that is no longer a 500 exception and 500 error message to Sentry
image

Image of production logs in Graylog (sans the _start_time, just ignore that field)
image


config/initializers/lograge.rb
```diff
@@ -0,0 +1,18 @@
...
+  config.lograge.formatter = Lograge::Formatters::Graylog2.new
+  if Rails.env.production?
+    config.log_tags = [ :remote_ip ]
+  end
+  config.lograge.base_controller_class = 'ActionController::API'
+  config.lograge.custom_options = lambda do |event|
+    exceptions = %w(controller action format id)
+    now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    {
+      start_time: (now - event.time).seconds.ago,
```
jpslav: Lets make this timestamp formatted like the one in brackets (i.e. "2019-06-06T22:..." instead of "Tue, 06 Aug...")

Actually... since the one in brackets is already prepended to request logs (and appears to be the same timestamp) I think we can remove this?

app/controllers/api/v0/search_controller.rb
```diff
@@ -78,6 +78,8 @@ def search
     response.overall_took = ((Time.now - started_at)*1000).round
 
     render json: response, status: :ok
+  rescue Elasticsearch::Transport::Transport::Errors::NotFound
+    render json: 'The specified resource was not found', status: 404
```
jpslav: If instead of rescuing directly here you let the exception flow out but use the `rescue_from_unless_local` at the top of this class, we can make sure to have standard error JSON and also let these errors explode in local development

app/services/monitor_index_jobs.rb
```diff
@@ -38,18 +38,20 @@ def notify_sentry_for_dead_letter_queue
...
-        Raven.capture_exception(ex)
-        log_error("dead job error #{ex.message} process #{dead_job.inspect}")
+        Raven.capture_exception(ex, :extra => dead_job.inspect)
+        Rails.logger.error("dead job error #{ex.message} process #{dead_job.inspect}")
       end
     end
   end
 
   def reset_desired_capacity_if_needed
+    return unless Rails.env.production?
```
jpslav: Why is this here?  This service seems to do a lot of other things that are AWS-only, so unclear why this method is different.

app/services/monitor_index_jobs.rb
```diff
@@ -38,18 +38,20 @@ def notify_sentry_for_dead_letter_queue
 
         @processed_from_dead += 1
       rescue => ex
-        Raven.capture_exception(ex)
-        log_error("dead job error #{ex.message} process #{dead_job.inspect}")
```
jpslav: We don't want the `MonitorIndexJobs:` prefix on the log?

app/services/work_index_jobs.rb
```diff
@@ -40,28 +41,27 @@ def call
...
-      enqueue_done_job(job: job,
-                       status: DoneIndexJob::STATUS_HTTP_ERROR)
+      handle_error(exception: ex, job: job, status: DoneIndexJob::STATUS_HTTP_ERROR)
     rescue => ex
-      Raven.capture_exception(ex)
-      Rails.logger.error("Exception occurred on #{job.to_json}. #{ex.message}")
-      enqueue_done_job(job: job,
-                       status: DoneIndexJob::STATUS_OTHER_ERROR,
-                       message: ex.message)
+      handle_error(exception: ex, job: job, status: DoneIndexJob::STATUS_OTHER_ERROR)
```
jpslav: good man :-)

jpslav (changes requested):

app/controllers/api/v0/search_controller.rb
steveburkett: thought we'd want local behaving like production here....

jpslav: talking about how Rails typically shows exceptions in development instead of rescuing them and showing error pages

jpslav (approved):

PR #43, opened 2019-08-12
load testing 

steveburkett:
A few changes resulting from load testing. 

1. Extract the book documents from the strategy so that this part is more SRP and extensible. 
1. Increase concurrency in puma from 5 to 10
1. Use the newest aws-ruby to get the new set_desired_capacity functionality.

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/214

image


app/services/books/indexing_strategies/i1/strategy.rb
```diff
@@ -29,50 +24,46 @@ def index_metadata
...
+
+      log_info("Creating index #{index_name} with #{documents.count} documents")
+
+      documents.each {|document| index_document(document: document) }
     end
 
-    def total_number_of_elements_to_index(book:)
-      book.root_book_part.pages.map{|page| page.elements(element_classes: DESIRED_ELEMENTS_TO_DOCUMENTS.keys).count }.sum
+    def total_number_of_documents_to_index(book:)
+      BookDocumentIterator.new(book: book).docs.count
```
jpslav: `BookDocs`?

app/services/load_testing_urls_writer.rb
```diff
@@ -0,0 +1,43 @@
+class LoadTestingUrlsWriter
+  LOAD_URLS_FILE_NAME = "load_urls.txt"
+
+  def initialize(env:, num_different_books: 3, urls_per_book: 3)
```
jpslav: leave num books nil and use all?

app/services/load_testing_urls_writer.rb
```diff
@@ -0,0 +1,43 @@
+class LoadTestingUrlsWriter
```
jpslav: Would prefer this not under `app` since it is for test purposes.  `lib` or other preferred

app/services/load_testing_urls_writer.rb
```diff
@@ -0,0 +1,43 @@
...
+  def find_terms_in_book_to_search_for(book_id)
+    book = OpenStax::Cnx::V1::Book.new(id: book_id)
+    random_docs = Books::IndexingStrategies::I1::BookDocs.new(book: book).docs.sample(@urls_per_book)
+    random_docs.each_with_object([]) do |doc, accum|
+      accum << doc.visible_content.split(' ').sort_by(&:length).reverse.first
+    end
+  end
+
+  def sample_some_books_ids
+    book_states = BookIndexState.live
```
jpslav: might just want ones that are created, this includes errors and creating

app/services/load_testing_urls_writer.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+  end
+
+  private
+
+  def write_urls(file, book_id, terms_to_use)
+    terms_to_use.each do |term|
+      stripped = term.gsub(/[^a-z0-9\-\s]/i, '').strip
+      url = "https://search-#{@env}.sandbox.openstax.org/open-search/api/v0/search?q=#{stripped}&books=#{book_id}&index_strategy=i1&search_strategy=s1"
```
jpslav: minor but don't need `/open-search/` in the URLs

app/services/load_testing_urls_writer.rb
```diff
@@ -0,0 +1,43 @@
...
+    @urls_per_book = urls_per_book
+    @env = env
+  end
+
+  def call
+    book_ids_to_use = sample_some_books_ids
+    File.open(LOAD_URLS_FILE_NAME, 'w') do |file|
+      book_ids_to_use.each do |book_id|
+        terms_to_use = find_terms_in_book_to_search_for(book_id)
+        write_urls(file, book_id, terms_to_use)
```
jpslav: Can we configure siege to take URLs in random order? if not this will write all URLs for one book, then the next, and so on, which won't be as realistic of a test.

app/services/load_testing_urls_writer.rb
```diff
@@ -0,0 +1,43 @@
...
+    terms_to_use.each do |term|
+      stripped = term.gsub(/[^a-z0-9\-\s]/i, '').strip
+      url = "https://search-#{@env}.sandbox.openstax.org/open-search/api/v0/search?q=#{stripped}&books=#{book_id}&index_strategy=i1&search_strategy=s1"
+      file.puts url
+    end
+  end
+
+  def find_terms_in_book_to_search_for(book_id)
+    book = OpenStax::Cnx::V1::Book.new(id: book_id)
+    random_docs = Books::IndexingStrategies::I1::BookDocs.new(book: book).docs.sample(@urls_per_book)
```
jpslav: why is the number of urls per book equal to the number of docs?

app/services/load_testing_urls_writer.rb
steveburkett: yes, siege has the -i flag that uses the URLs in random order

app/models/book_index_state.rb
```diff
@@ -73,6 +73,10 @@ def self.live
     all.reject{ |doc| doc.deleting? }
   end
 
+  def self.created
+    live.select{ |doc| doc.created? }
```
jpslav: Could be `all.select{...}`; small repeated work to reuse `live` here since the `reject` call in it is not useful here.

jpslav (approved):

PR #48, opened 2019-09-13
Separate out 500 http errors to reclassify as "OTHER_ERRORS"

steveburkett:
OTHER_ERRORS will be retried to index due to their dynamodb records being deleted, when the reenqueue job comes back around.

Non 500 http errors will be treated the same (they are retained in dynamodb to get the attention of the team to fix)

example Sentry error https://sentry.cnx.org/openstax/search/issues/7073/

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/501


app/services/work_index_jobs.rb
```diff
@@ -55,6 +59,11 @@ def call
 
   private
 
+  def error_500?(ex)
```
jpslav: `error_5xx?`

spec/spec_helper.rb
```diff
@@ -1,4 +1,12 @@
 require "rspec/json_expectations"
+
+require 'webmock/rspec'
+include WebMock::API
+# By default, allow webmock to connect.  In a specific
+# test to use, disallow webmock, stub_request, and then
+# allow_net_connect! in the test.
+WebMock.allow_net_connect!
```
jpslav: I think `WebMock` was included before due to VCR, so I'm not sure if this `allow_net_connect!` is needed?  And if it isn't, then we don't need to disable and re-enable it in the specs maybe?

jpslav (changes requested):

app/services/done_index_job.rb
```diff
@@ -47,8 +49,12 @@ def successful?
...
   end
 
-  def http_error?
-    status == STATUS_HTTP_ERROR
+  # For some http errors, we dont retry.
+  #
+  # e.g, for 404s, we'd like for the book to stay in the index
+  # state w/ the error so we can find out why it's not in cnx content.
+  def job_error_to_not_retry?
+    status == STATUS_HTTP_404_ERROR
```
jpslav: love it

app/services/work_index_jobs.rb
```diff
@@ -55,6 +55,20 @@ def call
 
   private
 
+  def handle_openstax_http_error(job, ex)
+    matches = ex.message.match(/(?<status_code>5\d\d)/)
```
jpslav:
```ruby
status_code = ex.message.match(/(\d\d\d)/)[1].to_i

status_error = 
  case status_code
  when 404
    DoneIndexJob::STATUS_HTTP_404_ERROR
  when 500..599
    DoneIndexJob::STATUS_HTTP_5XX_ERROR
  else
    DoneIndexJob::STATUS_HTTP_OTHER_ERROR
  end

handle_error(exception: ex, job: job, status: status_error)
```

spec/services/work_index_jobs_spec.rb
```diff
@@ -57,16 +57,80 @@
...
+        allow_any_instance_of(TodoJobsQueue).to receive(:read).and_return(job_with_exception)
+      end
+
+      it 'calls handle error with correct status' do
+        expect_any_instance_of(described_class)
+          .to receive(:handle_error)
+                .with(exception: anything, job: anything, status: DoneIndexJob::STATUS_OTHER_ERROR)
+        expect {
+          work_index_jobs.call
+        }.to_not raise_error(Exception)
```
jpslav: I think generally you can just say `to_not raise_error`

jpslav (approved):

PR #53, opened 2019-11-11
Make sure 500s get to sentry even when rescuing exceptions

jpslav:
Previously we were rescuing 500s in production and not sending them to sentry.  Oops.

### DIFF
--- a/lib/rescue_from_unless_local.rb
+++ b/lib/rescue_from_unless_local.rb
@@ -12,7 +12,9 @@ module RescueFromUnlessLocal
     # exceptions (good in development) or rescue exceptions (good in production).  Because
     # the test of the setting is done at the moment the exception is being handled, we
     # can change the value in specs to test both the exception and the rescuing behavior.
-    def rescue_from_unless_local(*klasses, with: nil, &block)
+    # If send_to_sentry is true and the exception is not already going to be reraised,
+    # the exception will be captured off to Sentry.
+    def rescue_from_unless_local(*klasses, with: nil, send_to_sentry: false, &block)
       rescue_from(*klasses, with: with) do |exception|
         # This new_block assignment borrowed from pieces of ActiveSupport::Rescuable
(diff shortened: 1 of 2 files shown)


lib/rescue_from_unless_local.rb
```diff
@@ -35,7 +37,12 @@ def rescue_from_unless_local(*klasses, with: nil, &block)
             end
           end
 
-        raise exception if Rails.application.config.consider_all_requests_local
+        if Rails.application.config.consider_all_requests_local
```
steveburkett: if one is local and wants to test the app sending to Raven, this local check needs to be commented out. that's ok?

lib/rescue_from_unless_local.rb
```diff
@@ -12,7 +12,9 @@ module ClassMethods
     # exceptions (good in development) or rescue exceptions (good in production).  Because
     # the test of the setting is done at the moment the exception is being handled, we
     # can change the value in specs to test both the exception and the rescuing behavior.
-    def rescue_from_unless_local(*klasses, with: nil, &block)
+    # If send_to_sentry is true and the exception is not already going to be reraised,
+    # the exception will be captured off to Sentry.
+    def rescue_from_unless_local(*klasses, with: nil, send_to_sentry: false, &block)
```
steveburkett: why is the default for send_to_sentry true for base_controller, but false here?

steveburkett (approved):

lib/rescue_from_unless_local.rb
jpslav: No you can just set this flag in `development.rb` to the value you want or you can stub it in the specs (we do such stubbing in highlights)

lib/rescue_from_unless_local.rb
jpslav: the one in the base controller isn't a default, its explicitly setting it to `true`.  Making the default in the `lib` be false makes this a non-breaking change.

PR #64, opened 2021-04-29
change info endpoint to have version stuff, move elasticsearch stuff …

TomWoodward:
…to a different endpoint


app/services/index_info.rb
```diff
@@ -74,4 +82,16 @@ def update_stat(index:, value_sym:, value:)
...
+  def env_name
+    ENV['ENV_NAME'] || 'Not set'
+  end
+
+  def ami_id
+    ENV['AMI_ID'] || 'Not set'
+  end
+
+  def git_sha
+    `git show --pretty=%H -q`&.chomp || 'Not set'
```
reedstrm, line 95: so, if deployment is not a git checkout (an export or tarball) this ends up as "not set"?

TomWoodward: yea i didn't invent this paradigm but apparently the ami does end up with the `.git` stuff

jpslav (changes requested):

jpslav (approved):

PR #66, opened 2021-05-12
Add support for KeyTerm

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1569

Requires / Based on https://github.com/openstax/cnx-ruby/pull/11


app/services/books/indexing_strategies/i1/key_term_document.rb
```diff
@@ -0,0 +1,18 @@
...
+            page_position: page_position,
+            page_id: page_id)
+    end
+
+    def self.element_type
+      "key_term"
+    end
+
+    def visible_content
+      [element.term, element.description]
```
TomWoodward, line 15: i was thinking the term would be the title, so that we can weight it higher if we want to. does this work though with returning multiple visible content items?

TomWoodward (changes requested):

app/services/books/indexing_strategies/i1/key_term_document.rb
PiotrKozlowski: Idk how to test it other than by writing tests, but `visibleContent` is an array according to generated types that frontend is using
I can change it to use `title`, but I would have to add support for `title` also in [here](https://github.com/openstax/open-search/blob/master/app/controllers/api/v0/swagger/models/search.rb#L10), right?
It seems like we only use `visible_content` and do nothing for `hidden_content` and `title` properties

PiotrKozlowski: When I was using mock version on the frontend with multiple visible elements I had to change how we calculate highlights because it was adding the length of `visibleContent` and title + desc for key term should be treated as one.
So maybe it would be better to use `title` overall

TomWoodward (approved):

PR #69, opened 2021-05-26
Do not index .os-teacher elements

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1559

TODO:
- [x] Update cnx-ruby: https://github.com/openstax/cnx-ruby/pull/14

In order to record new cassete I had to update webmock to version 3.5.0: https://github.com/bblimke/webmock/issues/786#issuecomment-450048667
I didn't include this change.
I've also tested locally that after updating cnx-ruby it is indeed working as expected.

### DIFF
--- a/app/services/books/indexing_strategies/i1/book_docs.rb
+++ b/app/services/books/indexing_strategies/i1/book_docs.rb
@@ -14,6 +14,8 @@ module Books::IndexingStrategies::I1
         @book.root_book_part.pages.each_with_object([]) do |page, accum|
           next if page.preface? || page.index?
 
+          page.remove_elements(xpath: "//*[contains(@class, 'os-teacher')]")
+
           page.
             elements(element_classes: DESIRED_ELEMENTS_TO_DOCUMENTS.keys).
             each_with_index { |element, page_position|
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,7 @@ gem 'sentry-raven'
 
 gem 'nokogiri'
 
-gem 'openstax_cnx', github: 'openstax/cnx-ruby', ref: '8bc389571'
+gem 'openstax_cnx', github: 'openstax/cnx-ruby', ref: '480b4285e'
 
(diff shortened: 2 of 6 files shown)


app/services/books/indexing_strategies/i1/book_docs.rb
```diff
@@ -15,7 +15,7 @@ def docs
           next if page.preface? || page.index?
 
           page.
-            elements(element_classes: DESIRED_ELEMENTS_TO_DOCUMENTS.keys).
+            elements(element_classes: DESIRED_ELEMENTS_TO_DOCUMENTS.keys, omit_by_xpath: "//*[contains(@class, 'os-teacher')]").
```
TomWoodward, line 18: i think it would be more clear and re-usable to have a `remove_elements` that we call before querying out `elements`

TomWoodward (changes requested):

app/services/books/indexing_strategies/i1/book_docs.rb
PiotrKozlowski: Should we remove them also by passing classes like for desired elements or is the usage of xpath okay here?
If we want to use classes it would be good to create them in `cnx-ruby`, but then when you want to remove a new element then you have to update 2 repos

PiotrKozlowski: I've added `remove_elements(xpath: string)`

TomWoodward (changes requested):
looks good. i merged the change in cnx_ruby, now the gemfile needs to be updated here right?

PiotrKozlowski:
@TomWoodward Yes. I'e updated it with https://github.com/openstax/cnx-ruby/commit/480b4285e14505606690efc29932fd28c692b9a5 and forced pushed the branch

TomWoodward (approved):

PR #70, opened 2021-05-28
Customize analyzer to match other quotation marks

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1468

We were using a **simple analyzer** which uses a lowercase tokenizer (see https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-simple-analyzer.html)

In order to match quotes, we have to add `char_filter` with proper mapping. I did it by overriding the default analyzer to our custom one.

Instead of using a `lowercase` tokenizer, I've used a `standard` one. As you can see [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html), the standard analyzer/tokenizer is producing different tokens:
```
POST _analyze
{
  "analyzer": "standard",
  "text": "The 2 QUICK Brown-Foxes jumped over the lazy dog's bone."
}

[ the, 2, quick, brown, foxes, jumped, over, the, lazy, dog's, bone ]
```
vs simple analyzer (lowercase tokenizer):
```
POST _analyze
{
  "analyzer": "simple",
  "text": "The 2 QUICK Brown-Foxes jumped over the lazy dog's bone."
}

[ the, quick, brown, foxes, jumped, over, the, lazy, dog, s, bone ]
```


app/services/books/indexing_strategies/i1/strategy.rb
```diff
@@ -56,7 +56,22 @@ def settings
...
+                tokenizer: "standard",
+                char_filter: [
+                  "quotes"
+                ]
+              }
+            },
+            char_filter: {
+              quotes: {
+                mappings: [
+                  "’=>'",
```
PiotrKozlowski, line 70: do we want to add more mappings here?

TomWoodward:
the way the `simple` tokenizer looks like its working in your description, where it splits the `dog` from the `s` looks like it should actually work, but the issue would be that it _doesnt_ do that for other quote types. is there an advantage to doing this in the new way where we actually tokenize the entire word including the quotes? i'm not sure its obvious to me which will produce better results

PiotrKozlowski:
@TomWoodward I don't have a strong argument for a standard analyzer. I was convinced be the docs that say it is the default one and it should work for most languages.
I've changed it back to the simple analyzer. Now there are 35 results for the same tests (increase from 25) and the highlights are a little bit different. [See the diff](https://github.com/openstax/open-search/pull/70/commits/5d66ce5f9b9fac0e27cb94bbb1c7e2b4fdf08188)

TomWoodward:
oh well the highlights going separately around the `s` is definitely less than ideal, thats a reason. if the docs also recommend the standard one i think its probably fine

PiotrKozlowski:
@TomWoodward I've reverted the previous commit

TomWoodward (approved):

PR #73, opened 2021-07-16
update search to be RAP pipeline aware

jpslav:
for: openstax/unified#1567

Makes search aware of RAP pipelines.  Legacy books continue to have an ID of the form `uuid@version`.  RAP books have an ID of the form `pipeline/uuid@version`.  Changes are included to pull each book from its appropriate archive URL (either "https://openstax.org/contents" or "https://openstax.org/apps/archive/#{pipeline}/contents".


Gemfile.lock
```diff
@@ -316,10 +316,10 @@ GEM
       activesupport (>= 3)
       railties (>= 3)
       yard (~> 0.9.11)
-    webmock (3.3.0)
+    webmock (3.13.0)
```
jpslav, line 319: Had to update this to make specs work in Ruby 2.6

spec/support/temp_aws_env.rb
```diff
@@ -126,16 +126,22 @@ def self.switch_to_temporary_credentials
...
-      ENV['AWS_MY_ARN'] = Aws::IAM::CurrentUser.new(region: "us-east-1").arn
-      filter_env_var('AWS_MY_ARN')
-
       sts_client = Aws::STS::Client.new(region: "us-east-1")
-      resp = sts_client.get_session_token({duration_seconds: 3600})
-      @@temporary_credentials = resp.credentials
+      identity = sts_client.get_caller_identity
+      already_in_temporary_role = identity.arn.match?(/:assumed-role\//)
+
+      if !already_in_temporary_role
```
jpslav, line 133: we didn't use to have aws-access and temporary credentials; if we are using aws-access skip the getting of new temporary credentials (AWS won't let us get temporary credentials from temporary credentials).

lib/bucket.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def file(key:)
+    BucketFile.new(key: key, region: @region, bucket_name: name)
+  end
+
+  def client
+    @client ||= ::Aws::S3::Client.new(region: @region)
+  end
+end
```
jpslav, line 22: s3 logic got a little more complicated so I wrote little `Bucket` and `BucketFile` classes to make `RexReleases` more readable.

app/services/books/index.rb
```diff
@@ -10,6 +10,9 @@ class Index
 
     DEFAULT_INDEXING_STRATEGY = IndexingStrategies::I1::Strategy
 
+    RAP_URL_BASE = "https://openstax.org/apps/archive"
+    LEGACY_URL = "https://openstax.org/contents"
```
jpslav, line 14: These were hardcoded before as `https://archive.cnx.org`, not sure if we want them to be configurable?

app/models/book.rb
```diff
@@ -0,0 +1,34 @@
+class Book < OpenStax::Cnx::V1::Book
+
+  RAP_URL_WITHOUT_PIPELINE = "https://openstax.org/apps/archive"
+  LEGACY_ARCHIVE_URL = "https://openstax.org/contents"
```
jpslav, line 4: These were hardcoded before as `https://archive.cnx.org`, not sure if we want them to be configurable?

jpslav:
Ok, I believe this works.  I've deployed it and it is indexing RAP books (tho I haven't formulated a search query on a deployed instance yet).  However there are a few issues:

1. Maybe due to the fact that pulling data from RAP is faster and less error-prone than pulling from legacy, a production deployment gets slammed a bit much by 30 workers.  Cluster goes red, elasticsearch instances die, etc.  We may want to consider switching to a smaller number of workers, like 15 or 20.  
2. The 30 workers are probably also overwhelming legacy archive.  I see 503s in Sentry and Graylog as it tries to pull JSON from legacy.  I tried to be smart and add an exponential backoff retry when we get a 503 in search, but then I realized that CloudFront is caching the 503s for a while.  For requests that go to legacy, we could add a nonsense query parameter at the end to force CloudFront to hit the origin again (legacy archive ignores params it doesn't know about).  But... if we are on the verge of not using legacy anymore maybe this isn't important?
3. When something goes wrong in an worker doing an indexing job, an exception is thrown and I believe the worker continues looking for new work, but not on the job it had been working on.  That job, which exists as an item in the "todo" SQS queue, must wait for its visibility timeout to expire -- that is currently 1 hour.  So anytime something explodes that job won't get worked on again for an hour.  We may want to either (1) more aggressively protect against known error conditions and retry or (2) intercept these errors briefly and then [zero out the job's visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#terminating-message-visibility-timeout) before continuing with raising and exploding (so that another worker can see it immediately).

TomWoodward:
1. i agree with adding something to manage the ES cluster load. we could also be more clever and actually check the cluster health and use that as a factor in scaling the worker group. just setting the max to something smaller would be easy though.
2. cloudfront shouldn't be caching error responses, actually in the docs it says `By default, when your origin returns an HTTP 4xx or 5xx status code, CloudFront caches these error responses for 10 seconds.`, but still i wouldn't think that would be an issue. i definitely don't think we should be bypassing cloudfront cache for legacy, because the cloudfront cache is the only thing that makes legacy stable enough to use.
3. this doesn't seem like we totally need to address it right now, but it sounds like a bug in the job handling? are you saying that the worker just drops failed jobs? i think there is supposed to be a way to mark them as failed, and we can just wrap the entire worker in a try/catch

jpslav:
Re (2), thanks for that, so I guess we could make the retries try after greater than 10 seconds.
Re (3) nevermind, the job is removed from the queue immediately, and in the event of an exception is written to a "done" queue.  The worker that processes the "done" queue effectively puts certain kinds of fixable exceptions back into the "todo" queue, so the visibility timeout thing only happens if a worker hangs, which should be very rare.

TomWoodward (approved):

PR #76, opened 2021-12-15
add support for archive override in release

TomWoodward:


lib/rex/release.rb
```diff
@@ -12,7 +12,10 @@ def initialize(id:, data:, config:)
 
     def books
       @books ||= @data["books"].map do |uuid, info|
-        Book.new(pipeline: config.pipeline, uuid: uuid, version: info["defaultVersion"])
+        pipeline = config.pipeline
+        pipeline = info['archiveOverride'].delete_prefix('/apps/archive/') if info.has_key?(:archiveOverride)
```
Dantemss, line 16: Normally I'd use a ternary operator but really what you have here is fine.

Dantemss (approved):
Looks good

lib/rex/release.rb
Dantemss: Accessing info with both a String 'archiveOverride' and a symbol :archiveOverride is normally a bit suspicious but it works because of the `.with_indifferent_access` call in the initializer.

`delete_prefix` is something I never used before and it's a scary name but the docs say it does not modify the String but rather create a copy so it's also fine.

TomWoodward: i was looking for a function that specifically removed a substring from the beginning, i found a few others that removed from anywhere in the string, this one seemed to do what i wanted.

i'm not clear on what places need to be symbols, i thought has_key wanted symbols all the time, but i guess i should have used symbols in both places?

PR #77, opened 2021-12-16
fix page archive url for github books

TomWoodward:
github books have page ids that end in an `@` with no version, like 
```
      {
        "id": "810e10c5-2e6b-47c1-aa36-a2645f3af52e@",
        "title": "<span data-type=\"\" itemprop=\"\" class=\"os-text\">Introduction</span>",
        "slug": "1-introduction"
      }
```


Dantemss (approved):
Looks good

PR #107, opened 2023-08-02
Update search

Dantemss:

Redoing the test cassettes was annoying.

I believe the extremely high number of deleted lines all comes from CNX-specific tests that I had to remove because those are gone. Likely there was an entire CNX book recorded there. I left the RAP versions in though.


app/controllers/api/v0/swagger/models/search.rb
```diff
@@ -76,29 +76,37 @@ module Api::V0::Swagger::Models::Search
...
+        key :required, [:value, :relation]
+        property :value do
+          key :type, :integer
+          key :readOnly, true
+          key :description, "The exact or approximate of hits"
+        end
+        property :relation do
+          key :type, :string
+          key :readOnly, true
+          key :description, "Modifier for value"
```
TomWoodward, line 108: are we changing the schema of the response here? this'll require a new client and rex updates to be compatible right?

TomWoodward (approved):
i'm not sure but it seems like there were changes to the api response format? if so we need to make corresponding rex changes in conjunction with this. or if at all possible it would be nice to make the response format backwards compatible

app/controllers/api/v0/swagger/models/search.rb
Dantemss: I added a commit to revert this change and added a translation in the controller so we can keep using v0

PR #108, opened 2023-08-15
Fixed the character filter

Dantemss:

For some reason the current char_filter causes a bizarre error where the AWS signature fails on the index create request. Maybe something in there really doesn't like the `=>` but I couldn't find info about this.
However, the pattern_replace type seems equivalent and doesn't cause this issue.


TomWoodward (approved):
I'm just trusting you on this one

PR #110, opened 2023-08-16
Fixed index deletion triggering when it shouldn't

Dantemss:

It's unclear why this bug did not affect elasticsearch.

in_demand was being set here: https://github.com/openstax/open-search/pull/110/files#diff-5482cda0b16b7ea9943dc675c7d6f64bc22453f5ea90444f6dc4cd7a1817c118R19

However, because index_states is a lazy loader, this change was not being propagated to it when it was accessed here: https://github.com/openstax/open-search/pull/110/files#diff-5482cda0b16b7ea9943dc675c7d6f64bc22453f5ea90444f6dc4cd7a1817c118R26

So all the indices were being deleted right after creation.

### DIFF
--- a/app/services/enqueue_index_jobs.rb
+++ b/app/services/enqueue_index_jobs.rb
@@ -23,7 +23,7 @@ class EnqueueIndexJobs
       end
     end
 
-    _, unneeded_book_indexings = index_states.partition(&:in_demand)
+    unneeded_book_indexings = index_states.reject(&:in_demand)
 
     unneeded_book_indexings.each do |unneeded_book_indexing|
       enqueue_delete_index_job(unneeded_book_indexing)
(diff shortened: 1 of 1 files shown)


app/services/enqueue_index_jobs.rb
```diff
@@ -23,7 +23,7 @@ def call
       end
     end
 
-    _, unneeded_book_indexings = index_states.partition(&:in_demand)
+    unneeded_book_indexings = index_states.reject(&:in_demand)
```
Dantemss, line 26: This is just cleanup and should have no effect.

app/services/enqueue_index_jobs.rb
```diff
@@ -44,7 +44,7 @@ def stats
   end
 
   def index_states
-    @index_states ||= BookIndexState.live
+    @index_states ||= BookIndexState.live.to_a
```
Dantemss, line 47: Changing index_states from a lazy loader to an array fixes the issue with in_demand being lost.

TomWoodward (approved):

PR #112, opened 2023-08-24
Added indexing strategies for books and pages based on ORN

Dantemss:
For: https://github.com/openstax/unified/issues/2523

- Book and page indices
- Added APIs to search these indices

Requires: https://github.com/openstax/openstax-resource-names/pull/13


TomWoodward (changes requested):
i think this is making indexes with only one book document in them, we want one index for all the books but i don't see any changes to index creation

Dantemss:
I'm thinking this is ready. I'm deploying to one of the releases to give it a test...

TomWoodward (approved):
looks ok to me but i'd like to see it working after the orn deployment goes out

Dantemss:
Waiting for the ORN fix to go out.

Dantemss:
While implementing something to use this in ORN I started thinking maybe we need to add a book id field to i2 and i3 (or all indices) and either add an s2 strategy that can filter by book id or modify s1.

ORN currently expects the book, page and element searches to be filterable by book id.

PR #115, opened 2023-09-26
Handle smart quotes and hyphens

Dantemss:
For: https://github.com/openstax/unified/issues/2613

- Handle smart quotes by converting them to normal quotes before searching
- Ignore mismatched quotes, dashes and em-dashes (consider them the same as whitespace)

### DIFF
--- a/app/services/books/search_strategies/s1/strategy.rb
+++ b/app/services/books/search_strategies/s1/strategy.rb
@@ -24,16 +24,17 @@ module Books::SearchStrategies::S1
 
     def fuzzify(query_string)
       # 1. Remove ~ and optional suffix from query (to prevent users setting their own fuzziness)
-      # 2. Split into array of unquoted words and quoted phrases
-      # 3. Add fuzziness to unquoted words only
-      # 4. Join array back into a string with spaces
+      # 2. Convert smart quotes to normal quotes
+      # 3. Split into array of unquoted words and quoted phrases
+      # 4. Fix mismatched quotes, add fuzziness to unquoted words only
+      # 5. Join array back into a string with spaces
       #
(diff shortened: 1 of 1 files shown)


TomWoodward (changes requested):
why is it necessary to close the mismatched quotes? i feel like this is going to lead to unexpected behavior if you just happen to have a quote in your search term (like maybe you copy pasted half a quote out of the content or something). would it be better to ignore mismatched quotes? what does it do now?

Dantemss:
What it does now is that mismatched quotes disable the fuzzy search, but they are not treated as phrases by OpenSearch so the words can be in any order. If you want I can do a gsub or something to remove the quotes completely so they are treated like unquoted words, or I can leave them as this kind of in-between behavior if you think it would be useful.

TomWoodward:
i'm not sure i'm totally understanding, but i would think that a mismatched quote should essentially be ignored

Dantemss:
Changed pattern so mismatched quotes are ignored:

```
irb(main):002:0> 'the quick brown fox'.scan(/[^\s"—–-]+|"[^"]*"/)
=> ["the", "quick", "brown", "fox"]
irb(main):003:0> 'the quick "brown fox'.scan(/[^\s"—–-]+|"[^"]*"/)
=> ["the", "quick", "brown", "fox"]
irb(main):004:0> 'the quick "brown fox"'.scan(/[^\s"—–-]+|"[^"]*"/)
=> ["the", "quick", "\"brown fox\""]
irb(main):005:0> 'the quick"brown fox'.scan(/[^\s"—–-]+|"[^"]*"/)
=> ["the", "quick", "brown", "fox"]
```

TomWoodward (approved):

PR #117, opened 2024-02-16
Added language-specific stemming

Dantemss:
For: https://github.com/openstax/unified/issues/2625

This will let you search for verb conjugations and plurals, for example: https://release-2757.sandbox.openstax.org/books/physics/pages/1-introduction?query=exert

### DIFF
--- a/app/services/books/search_strategies/s1/strategy.rb
+++ b/app/services/books/search_strategies/s1/strategy.rb
@@ -34,7 +34,7 @@ module Books::SearchStrategies::S1
       # Fuzziness values replicate the AUTO fuzziness of other OpenSearch query types
       # simple_query_string doesn't seem to support AUTO fuzziness
       query_string.gsub(/~[^\s"]*/, '').gsub(/“|”/, '"').scan(/[^\s"—–-]+|"[^"]*"/).map do |str|
-        next str if str.start_with?('"') || str.length <= 3
+        next str if str.start_with?('"') || str.length <= 2
 
         fuzziness = str.length <= 5 ? 1 : 2
         "#{str}~#{fuzziness}"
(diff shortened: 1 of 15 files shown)


Dantemss:
I tried adding code to reduce the instance size after indexing but this ended up causing the cluster to become unresponsive so I gave up

app/services/books/search_strategies/s1/strategy.rb
```diff
@@ -49,10 +49,11 @@ def search_body(query_string)
         query: {
           simple_query_string: {
             fields: %w(title contextTitle visible_content),
-            query: query_string,
-            flags: "PHRASE|WHITESPACE",
+            query: fuzzify(query_string),
```
TomWoodward, line 52: did you mention that maybe the stuff above at `query_string.gsub` is no longer necessary anymore?

TomWoodward (approved):

app/services/books/search_strategies/s1/strategy.rb
Dantemss: Erika wanted to try fuzzy search again as well, but this time with the fuzzy_prefix_length set to 3 so mistakes in the first 3 characters are ignored

TomWoodward: right but i meant for the quote replacement

Dantemss: The gsub stuff above is part of the fuzzy search so it's necessary. The thing that can be removed is the smart quotes char filter for English specifically, since I added asciifolding to map all non-ascii characters to ascii for English (but not for Spanish/Polish since they use non-ascii characters)

TomWoodward (approved):

PR #118, opened 2024-04-05
DISCO-104 Use multi_match instead of simple_query_string

Dantemss:
Per Erika's decision, I'm changing this PR to make search use multi_match instead of simple_query_string, which removes some user control over the query but should give us both stemming and fuzziness.

### DIFF
--- a/spec/services/books/search_strategies/s1/strategy_spec.rb
+++ b/spec/services/books/search_strategies/s1/strategy_spec.rb
@@ -121,8 +121,8 @@ RSpec.describe Books::SearchStrategies::S1::Strategy , type: :request, api: :v0,
 
     it 'applies fuzzy modifiers to unquoted words outside of parentheses only' do
       expect(search_strategy.send(
-        :fuzzify, 'a humongous word "in quotes" (and parentheses) sometimes(without)"spaces"'
-      )).to eq 'a humongous~2 word~1 "in quotes" (and parentheses) sometimes~2 (without) "spaces"'
+        :fuzzify, 'a humongous word "in quotes" (and parentheses) sometimes"(without)"("spaces")'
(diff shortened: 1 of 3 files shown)


spec/services/books/search_strategies/s1/strategy_spec.rb
```diff
@@ -115,4 +115,14 @@
...
     end
   end
+
+  context 'fuzzify' do
+    let(:index_names) { [index_name] }
+
+    it 'applies fuzzy modifiers to unquoted words outside of parentheses only' do
+      expect(search_strategy.send(
+        :fuzzify, 'a humongous word "in quotes" (and parentheses) sometimes"(without)"("spaces")'
+      )).to eq 'a humongous~2 word~1 "in quotes" (and parentheses) sometimes~2 "(without)" ("spaces")'
```
TomWoodward, line 125: this seems to be resulting in words within parenthesis not being fuzzy matched, and i'm not sure thats really what we mean to do. i'm not sure what the intent is here though. if we want the parenthesis to be used as the elastic search [precedence grouping](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax) then i think we need to find a way to make the words within continue to be fuzzy, maybe add a space around the `(` and `)` symbols would help?

if the issue is that we're unable to search for text excerpts containing literal parens characters, maybe we need to auto escape them or disable the precedence feature?

TomWoodward (changes requested):

spec/services/books/search_strategies/s1/strategy_spec.rb
Dantemss: Well, I can give adding spaces after ( and before ) a try...

TomWoodward (approved):

